PR #19, opened 2021-01-14
add event/{id} endpoint

### LINKED ISSUE #12: GET /event/{id}
- [Contract](https://pact.serlo.org/pacts/provider/serlo.org/consumer/api.serlo.org/latest#fetch_data_of_event_with_id_2115_given_event_2115_is_of_type_CreateEntityLinkNotificationEvent)

elbotho:
a bit less succinct as the other ones :)
see comments in code

(closes #12)


elbotho:
> Probably one of the places we should refactor sooner rather than later

Would you suggest a file for each type or other changes?

PR #27, opened 2021-01-15
POST /set-uuid-state

inyono:


inyono:
The database layer side of /POST set-uuid-state should be ready now. Needs some work in api.serlo.org beforehand though:
- [ ] Adapt contract test to new request body & return type.
- [ ] Check permissions beforehand.

PR #28, opened 2021-01-17
refactor /event

elbotho:
#### Status
`LIMIT=100 yarn test event.js` successful
(same two failures in legacy as last time)

---

#### Notes
- moved event types into files (was just too much for a single file, even if we have a bit more code this way)
- implementations are not async, all fetching happens in `event.rs`
- simplified queries a lot (expl. below)

The first query gets one or zero parameter ids (without checking the name etc.) which works for all but three types.
`entity/revision/checkout`, `entity/revision/reject` and `taxonomy/term/parent/change` need data from two parameter fields that is provided by `fetch_parameter_string(` and `fetch_parameter_uuid_id(`.
These second queries _could_ be joined/optimised, but I prefer the clarity.


inyono:
Thanks for the PR, has some nice improvements. I have some further suggestions, see #29. Let me know if we should discuss that in detail or you disagree with some of the points.

PR #29, opened 2021-01-19
refactor /event: further ideas

inyono:
Adds some further improvement ideas for #28.

### DIFF
--- a/src/event/model/checkout_revision.rs
+++ b/src/event/model/checkout_revision.rs
@@ -1,33 +1,15 @@
-use crate::event::model::CommonEventData;
+use super::event::AbstractEvent;
 use serde::Serialize;
 
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CheckoutRevision {
+    #[serde(flatten)]
+    pub abstract_event: AbstractEvent,
+
     #[serde(rename(serialize = "__typename"))]
     pub __typename: String,
-    pub id: i32,
-    pub instance: String,
-    pub date: String,
-    pub object_id: i32,
-    pub actor_id: i32,
     pub repository_id: i32,
     pub revision_id: i32,
     pub reason: String,
-}
-
-impl CheckoutRevision {
-    pub fn build(data: CommonEventData, repository_id: i32, reason: String) -> CheckoutRevision {
-        CheckoutRevision {
-            __typename: "CheckoutRevisionNotificationEvent".to_string(),
-            id: data.id,
-            instance: data.instance,
-            date: data.date,
-            object_id: data.uuid_id,
-            actor_id: data.actor_id,
-            repository_id,
-            revision_id: data.uuid_id,
-            reason,
-        }
-    }
 }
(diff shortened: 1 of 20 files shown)


src/event/model/checkout_revision.rs
```diff
@@ -1,33 +1,15 @@
-use crate::event::model::CommonEventData;
+use super::event::AbstractEvent;
 use serde::Serialize;
 
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CheckoutRevision {
+    #[serde(flatten)]
+    pub abstract_event: AbstractEvent,
+
```
inyono, line 9: This allows you to make the common fields explicit with a new struct, so:
- No copy-paste of common fields
- Less need for a dedicated constructor for each type using `CommonEventData` since you can pass `AbstractEvent` and use it in one go.

src/event/model/event.rs
```diff
@@ -91,9 +110,19 @@ impl Event {
...
-                Event::CheckoutRevision(CheckoutRevision::build(e, repository_id.unwrap(), reason))
+                let revision_id = abstract_event.object_id;
+                Event::CheckoutRevision(CheckoutRevision {
+                    abstract_event,
+
+                    __typename: String::from("CheckoutRevisionNotificationEvent"),
+                    repository_id,
+                    revision_id,
+                    reason,
+                })
```
inyono, line 125: It probably makes sense to outsource that into `checkout_revision`. If you do that, though, I'd prefer the pattern we used for the uuid model instead, i.e. instead of fetching the type-specific data here, let the type handle those themselves. So, for example:
```rust
impl CheckoutRevision {
    // new is the idiomatic name for a constructor-like function
    pub async fn new(abstract_event: AbstractEvent, pool) -> Result<Self> {
        let reason = Event::fetch_parameter_string(abstract_event.id, "reason", &pool).await;
    }
}
```

src/event/model/event.rs
```diff
@@ -91,9 +110,19 @@ impl Event {
...
+                let repository_id = fetch_parameter_uuid_id(e.id, "repository", &pool)
+                    .await
+                    .unwrap();
                 let reason = fetch_parameter_string(e.id, "reason", &pool).await;
-                Event::CheckoutRevision(CheckoutRevision::build(e, repository_id.unwrap(), reason))
+                let revision_id = abstract_event.object_id;
+                Event::CheckoutRevision(CheckoutRevision {
+                    abstract_event,
+
+                    __typename: String::from("CheckoutRevisionNotificationEvent"),
```
inyono, line 121: Would probably be even cleaner using Valentin's pattern:
- Enum https://github.com/serlo/serlo.org-database-layer/pull/23/files#diff-22cee5576a81bdb8d679e1d0f63bf4ef19552e831836d80ecf548213c881883cR47
- Parse from string https://github.com/serlo/serlo.org-database-layer/pull/23/files#diff-b6f1f849ff97d9bdafb07928931bdd1404acf4b0730e8a39ed8efbd75c52f3eaR32

inyono: This would also allow to move the `__typename` into `AbstractEvent`.

elbotho:
@inyono thanks for the suggestions!
I changed the code accordingly. 

Added two hidden fields to `AbstractEvent` to simplify some code, is that alright?
 ```
#[serde(skip)]
pub parameter_uuid_id: i32,
#[serde(skip)]
pub name: String,
```

PR #40, opened 2021-01-21
Add basic test structure

Lodifice:
For now we only test pact 1, but this should be sufficient for adding more tests. I'll think about testing corner cases (what we talked about yesterday) and maybe factoring out a setup function for the tests later.

Interestingly we got rid off tokio as a dependency and the tests woked just fine without runtime errors. No magic was necessary. Maybe there was a version bump of actix and they fixed it upstream?


inyono (approved):
Nice 🎉

PR #41, opened 2021-01-22
Use transactions in unit tests

inyono:


src/event/model/set_uuid_state.rs
```diff
@@ -22,3 +26,96 @@ impl From<AbstractEvent> for SetUuidState {
...
+        if let Event::SetUuidState(set_uuid_state_event) = event {
+            let persisted_event =
+                Event::fetch(set_uuid_state_event.abstract_event.id, &mut transaction)
+                    .await
+                    .unwrap();
+        // assert_eq!(persisted_event, event);
+        } else {
+            panic!("Wrong event type");
+        }
+    }
```
inyono, line 120: Since we don't `transaction.commit` here, the whole transaction while be rolled back, so tests don't change the database.

src/event/model/set_uuid_state.rs
```diff
@@ -22,3 +26,96 @@ impl From<AbstractEvent> for SetUuidState {
...
+            .fetch_one(&mut transaction)
+            .await
+            .map_err(|inner| EventError::DatabaseError { inner })?;
+        let event = Event::fetch(value.id as i32, &mut transaction).await?;
+        transaction
+            .commit()
+            .await
+            .map_err(|inner| EventError::DatabaseError { inner })?;
+
+        Ok(event)
```
inyono, line 93: This whole function is a transaction (or possibly a sub transaction when the passed executor is already a transaction like in the tests). So: the mutations and the concluding query will only be executed as a group.

src/event/model/event.rs
```diff
@@ -46,7 +46,14 @@ pub enum Event {
...
 impl Event {
-    pub async fn fetch(id: i32, pool: &MySqlPool) -> Result<Event, EventError> {
+    pub async fn fetch<'a, E>(id: i32, executor: E) -> Result<Event, EventError>
+    where
+        E: Executor<'a>,
+    {
+        let mut transaction = executor
+            .begin()
+            .await
+            .map_err(|inner| EventError::DatabaseError { inner })?;
```
inyono, line 56: Queries use the same trait (so that we can use queries from mutations). We don't need to roll back (or commit) the transaction here since it didn't change anything in the database. Interesting side effect: this guarantees that `Event::fetch` doesn't modify the database at all.

src/event/model/event.rs
```diff
@@ -87,10 +96,12 @@ impl Event {
...
         )
-        .fetch_all(pool);
+        .fetch_all(&mut transaction)
+        .await
+        .map_err(|inner| EventError::DatabaseError { inner })?;
 
-        let (string_parameters, uuid_parameters) = try_join!(string_parameters, uuid_parameters)
-            .map_err(|inner| EventError::DatabaseError { inner })?;
+        // let (string_parameters, uuid_parameters) = try_join!(string_parameters, uuid_parameters)
+        //     .map_err(|inner| EventError::DatabaseError { inner })?;
```
inyono, line 104: Open TODO: need to review whether we can still use `join` with this pattern (so that stuff that can be executed in parallel will be executed in parallel). Problem here at the moment is the mutable reference...

inyono (commented):
Some WIP stuff for mutations, see comments for some guidance on what's happening here.

inyono:
Superseded by #45.

Issue #42, opened 2021-01-25
Alias of revision shall be /entity/repository/compare/:id/:revisionId


kulla:
As discussed in the today session: The alias of a revision shall be set to /entity/repository/compare/:id/:revisionId

inyono:
FYI: this hasn't been deployed yet but will be part of the next release.

PR #52, opened 2021-01-27
change entity revision alias

elbotho:
(for #42)

or would you prefer something like:
``` rust
            alias: format_alias(
                Some("/entity/repository/compare/"),
                id,
                Some(&format!("{}", revision.repository_id)),
            ),
```


inyono (approved):
Thanks, that's good enough for now. We probably want to create a new type for Aliases at some point, though.

PR #53, opened 2021-01-28
POST /set-notification-state

### LINKED ISSUE #48: POST /set-notification-state
[Contract](https://pact.serlo.org/pacts/provider/serlo.org/consumer/api.serlo.org/latest#set_state_of_notification_with_id_9_given_there_exists_a_notification_with_id_9_for_user_with_id_1)

elbotho:
closes #48

Basically copy/pasted 90% from `set-uuid-state`.


__tests__/set-notification-state.js
```diff
@@ -0,0 +1,17 @@
...
+    body: JSON.stringify(body),
+    headers: { 'Content-Type': 'application/json' },
+  })
+  console.log(response)
+  try {
+    const data = await response.json()
+  } catch (e) {
+    console.log(e)
+  }
+})
```
inyono, line 17: Not needed.

src/notifications/model.rs
```diff
@@ -183,13 +183,194 @@ impl Notifications {
...
+                r#"
+                    UPDATE notification
+                        SET seen = ?
+                        WHERE id = ?
+                "#,
+                seen,
+                id
+            )
+            .execute(&mut transaction)
+            .await?;
```
inyono, line 254: We probably don't need the `SELECT` beforehand. I initially did that because I wanted to track which rows did change but didn't need it later. We might even be able to do this in a single statement.

src/notifications/model.rs
```diff
@@ -183,13 +183,194 @@ impl Notifications {
...
+        }
+
+        transaction.commit().await?;
+
+        Ok(Notifications::fetch_via_transaction(payload.user_id, pool).await?)
```
inyono, line 259: We won't need the notifications as a return value since we can calculate the cache ourselves in api.serlo.org after the mutation (similarly to what I did with setUuidState).

src/notifications/model.rs
```diff
@@ -183,13 +183,194 @@ impl Notifications {
...
+        // Verify that the object was set to read (eqals seen = true).
+        let uuid = sqlx::query!(r#"SELECT seen FROM notification WHERE id = ?"#, 6522)
+            .fetch_one(&mut transaction)
+            .await
+            .unwrap();
+        assert!(uuid.seen != 0);
+    }
+
+    #[actix_rt::test]
+    async fn set_notification_state_single_id_no_change_if_already_set() {
```
inyono, line 314:
```suggestion
    async fn set_notification_state_single_id() {
```

inyono (changes requested):
Thanks. Noted some things.

src/notifications/model.rs
elbotho: **But the pact man! The pact! 🙀** 

(okay, simpler this way.)

src/notifications/model.rs
elbotho: set_notification_state_single_id already exists, so basically remove this test?

inyono: Sorry, I misunderstood your test name then. Maybe `set_notification_state_read` and `set_notification_state_unread` are better names for the two single id cases?

inyono: What my confusion was: "if_already_set" implies to me that you test what happens when you set an already seen notification to seen. But that's not what the test is doing. Instead, you have two tests, one for the unread case, and one for the seen case (which is good)

elbotho: ^^
okay. 

>  "if_already_set" implies to me that you test what happens when you set an already seen notification to seen

That's exactly what I was trying to archive, but maybe I got it wrong?

Notification 6522 is already unread before mutation
I try to set it to unread
and I check if it's still unread after mutation ( `assert!(uuid.seen == 0);` )

elbotho: If it's right, I'll add an comment to explain.

Related: Should I add seperate tests for read/unread?

inyono: Well, that test doesn't make much sense, then, in my opinion. It would only test if we didn't do some weird toggle logic (which would be inefficient anyways since we needed to select the data previously). So I'd remove that test and rather have a test for unread / read each.

elbotho: check 👍

elbotho: tbh. the single id test does read, the multiple does unread, that's probably enough already?

inyono: Still would add two tests for that. Single id is the "easy" case where we test the interesting edge cases (so unread/read in this example). And multiple ids is a test that checks if this also works when passing multiple ids (w/o repeating the edge cases again). This makes things easier when stuff breaks (because we'd have more information about what exactly broke).

elbotho:
Thanks for the review and the test philosophy explanation :)

inyono (approved):

PR #56, opened 2021-01-28
POST /thread/set-archive

### LINKED ISSUE #46: POST /thread/set-archive
[Contract](https://pact.serlo.org/pacts/provider/serlo.org/consumer/api.serlo.org/latest#set_%22archived%22_state_of_a_thread_given_there_exists_a_thread_with_a_first_comment_with_an_id_of_100_and_user_with_id_1_is_authenticated)

elbotho:
(closes #46)


src/event/model/event_type.rs
```diff
@@ -7,7 +7,7 @@ pub enum EventType {
     #[serde(rename(
         serialize = "SetThreadStateNotificationEvent",
         deserialize = "discussion/comment/archive",
-        deserialize = "discussion/comment/restore"
+        deserialize = "discussion/restore"
```
inyono, line 10: @inyono Review the actual semantics of the types. Might be that we don't trigger an event for all possible combinations of threads.

src/event/model/event_type.rs
```diff
@@ -88,7 +88,7 @@ pub enum EventType {
     SetUuidState,
     #[serde(rename(
         serialize = "UnsupportedNotificationEvent",
-        deserialize = "discussion/restore"
+        deserialize = "discussion/comment/restore"
```
inyono, line 91: This is probably not needed anymore, right?

elbotho: to be honest idk: it was added for compatibility with legacy.

inyono (approved):
Thanks!

PR #61, opened 2021-01-29
POST /thread/comment-thread

elbotho:
----
**Actual queries legacy sends when adding comment.** 

Maybe helpful?
``` sql
START TRANSACTION

INSERT INTO uuid (trashed, discriminator) 
    VALUES (0, 'comment')

INSERT INTO comment ( id , date , archived , title , content , uuid_id , parent_id , author_id , instance_id )
    VALUES ( 35610 , '2021-01-29 02:26:19' , 0 , NULL , 'XANTWORTX' , NULL , 35609 , 1 , 1 )

INSERT INTO event_log ( date , actor_id , event_id , uuid_id , instance_id )
    VALUES ( '2021-01-29 02:26:20' , 1 , 9 , 35610 , 1 )
    
INSERT INTO event_parameter ( log_id , name_id )
    VALUES ( 86593 , 5 )
    
INSERT INTO event_parameter_uuid ( event_parameter_id , uuid_id )
    VALUES ( 70043 , 35609 )

COMMIT
```

### DIFF
--- a/src/subscriptions/model.rs
+++ b/src/subscriptions/model.rs
@@ -1,7 +1,9 @@
-use crate::database::Executor;
 use serde::Serialize;
 use sqlx::MySqlPool;
 use thiserror::Error;
+
+use crate::database::Executor;
+use crate::datetime::DateTime;
 
 pub struct Subscriptions(pub Vec<Subscription>);
 
@@ -76,6 +78,38 @@ impl Subscriptions {
 
         Ok(Subscriptions(subscriptions))
     }
+
+    pub async fn create_subscription<'a, E>(
+        object_id: i32,
+        user_id: i32,
+        send_email: bool,
+        executor: E,
+    ) -> Result<(), SubscriptionsError>
+    where
+        E: Executor<'a>,
+    {
+        let mut transaction = executor.begin().await?;
+        sqlx::query!(
+            r#"
+            INSERT INTO subscription (uuid_id, user_id, notify_mailman, date)
+                SELECT ?, ?, ?, ?
+                WHERE NOT EXISTS
+                (SELECT id FROM subscription WHERE uuid_id = ? AND user_id = ?)
+            "#,
+            object_id,
+            user_id,
+            send_email,
+            DateTime::now(),
+            object_id,
+            user_id,
+        )
+        .execute(&mut transaction)
+        .await?;
+
+        transaction.commit().await?;
+
+        Ok(())
+    }
 }
 
 #[derive(Serialize)]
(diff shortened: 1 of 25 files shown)


inyono:
> Everything's prepared for the real magic 😉

Is that a "You can take over the PR now" 😁

inyono:
Depends on #62, #63

inyono:
FYI: one of the reasons the CI fails is because `sqlx-data.json` is not up-to-date anymore (because you added / changed queries). Can you tell me if/how that is impacting your local dev process? Or does SQLX just ignore that with `cargo run`?

elbotho:
>  you tell me if/how that is impacting your local dev process?

Jup, no effect on `cargo test` and `cargo run` as far as I can tell.


Okay! #62 and #63 are added. 

**Todos:**
- figure out why comment `INSERT` query fails with `Err(DatabaseError { inner: RowNotFound })` even though the exact steps work manually with SQL. `cargo test thread` to run debug test.
- if that works, add test for successful cases.
- Check: Is the subscription to the thread or the comment?
- ?

elbotho:
BTW: how can I update `sqlx-data.json` or make the checks pass again?

src/subscriptions/model.rs
```diff
@@ -78,6 +82,35 @@ impl Subscriptions {
...
+    where
+        E: Executor<'a>,
+    {
+        let mut transaction = executor.begin().await?;
+        sqlx::query!(
+            r#"
+            INSERT INTO subscription (uuid_id, user_id, notify_mailman, date)
+                SELECT ?, ?, ?, ?
+                WHERE NOT EXISTS
+                (SELECT id FROM subscription WHERE uuid_id = ? AND user_id = ?)
```
inyono, line 96: This doesn't have the correct behavior yet. If the subscription already exists, we should still update `notify_mailman`.

src/threads/model.rs
```diff
@@ -135,14 +137,175 @@ impl Threads {
...
+    user_id: i32,
+}
+
+#[derive(Error, Debug)]
+pub enum ThreadCommentThreadError {
+    #[error("Comment could not be saved because of a database error: {inner:?}.")]
+    DatabaseError { inner: sqlx::Error },
+    #[error("Comment could not be saved because thread is archived")]
+    ThreadArchivedError,
+    #[error("Comment could not be saved because of a database error: {inner:?}.")]
```
inyono, line 154: Typo

src/threads/model.rs
```diff
@@ -135,14 +137,175 @@ impl Threads {
...
+            payload.user_id,
+            thread.instance_id,
+        )
+        .save(&mut transaction)
+        .await?;
+
+        let subscription = Subscription {
+            object_id: payload.thread_id,
+            user_id: payload.user_id,
+            send_email: true,
```
inyono, line 238: @inyono: Review that. Although I wouldn't be surprised if that is the behavior of legacy ^^'

src/threads/model.rs
```diff
@@ -135,14 +137,175 @@ impl Threads {
...
+
+        let comment = sqlx::query!(
+            r#"
+                SELECT content, author_id FROM comment WHERE parent_id = ?
+            "#,
+            17774
+        )
+        .fetch_one(&mut transaction)
+        .await
+        .unwrap();
```
inyono, line 285: we can use `Uuid` here.

src/threads/routes.rs
```diff
@@ -40,7 +43,32 @@ async fn set_archive(
...
+            println!("/thread/comment-thread: {:?}", e);
+            match e {
+                ThreadCommentThreadError::DatabaseError { .. } => {
+                    HttpResponse::InternalServerError().json(None::<String>)
+                }
+                ThreadCommentThreadError::ThreadArchivedError { .. } => {
+                    HttpResponse::InternalServerError().json(None::<String>)
+                }
+                ThreadCommentThreadError::EventError { .. } => {
+                    HttpResponse::InternalServerError().json(None::<String>)
```
inyono, line 63: These shouldn't have a response body.

inyono (changes requested):

src/threads/model.rs
inyono: Also the format doesn't follow our existing errors.

src/threads/model.rs
inyono: Okay. This should actually be a param in the payload (if you want to subscribe and if you want to get emails).

inyono: https://github.com/serlo/api.serlo.org/issues/234

inyono:
> BTW: how can I update `sqlx-data.json` or make the checks pass again?

- `cargo install sqlx-cli` once and then
- `cargo sqlx prepare -- --lib` to update sqlx-data.json

I'll update the README soon but wanted to have a feeling how this thing works before that ;)

inyono (approved):

PR #71, opened 2021-02-05
SubscriptionChangeMutation

elbotho:
Draft for a mutation to add, update and remove subscriptions.

Notes:
- Did not add a route, since we likely change to messages
- Feel free to change the naming (and everything ;)
- slightly preferred to have only one message for all actions, but could also be split up (would maybe split it up for the graphql mutations?)

### DIFF
--- a/sqlx-data.json
+++ b/sqlx-data.json
@@ -1065,6 +1065,16 @@
       "nullable": []
     }
   },
+  "72988b124c9f33d66c499a89625eef7388b981b0428ab123bba4808b11651b41": {
+    "query": "DELETE FROM subscription WHERE uuid_id = ? AND user_id = ?",
+    "describe": {
+      "columns": [],
+      "parameters": {
+        "Right": 2
+      },
+      "nullable": []
+    }
+  },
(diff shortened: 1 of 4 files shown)


inyono:
What are the planned mutations in api.serlo.org?

elbotho:
Idea:
Maybe just a `setSubsriptionState` would be most consistent and simple:
With an input of `{id, subscribe, sendEmail}`
Otherwise we would need a delete, add and change mutation, which seems a bit much for a simple thing that's likely to change in somewhat near the future.

What do you think?

inyono:
Okay. I'm fine with either
```graphql
mutation {
  uuid {
    setSubscribe(input: $input)
  }
}
```
(with https://github.com/serlo/api.serlo.org/issues/101 in mind) or
```graphql
mutation {
  subscription {
    set(input: $input)
  }
}
```
Your pick, @elbotho 😉

elbotho:
In this case, let's go with the second option.

inyono (approved):

elbotho:
✨  Thanks!

PR #73, opened 2021-02-06
add send_email to SubscriptionByUser

elbotho:
(for https://github.com/serlo/api.serlo.org/issues/231)


inyono (approved):
A little bit surprised that the contract tests aren't failing (seems like additional keys are okay when pact compares arrays) 🤷‍♂️

PR #89, opened 2021-05-06
course_revision: replace content w/ description

### LINKED ISSUE #88: Review CourseRevision
Seems like Course incorrectly query a non-existing "content" field instead of "description". We should review that.

elbotho:
Hm... that should be it? 
Seems a bit easy ;)
Tested with 30700, could not find a `CourseRevision` with description in the local data..

closes #88

(also some updates to Cargo.lock slipped in)

### DIFF
--- a/src/uuid/model/entity_revision/course_revision.rs
+++ b/src/uuid/model/entity_revision/course_revision.rs
@@ -6,21 +6,21 @@ use super::abstract_entity_revision::AbstractEntityRevision;
 #[serde(rename_all = "camelCase")]
 pub struct CourseRevision {
     title: String,
-    content: String,
+    description: String,
     meta_description: String,
 }
 
 impl From<&AbstractEntityRevision> for CourseRevision {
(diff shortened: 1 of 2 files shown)


src/uuid/model/entity_revision/course_revision.rs
```diff
@@ -6,21 +6,21 @@ use super::abstract_entity_revision::AbstractEntityRevision;
 #[serde(rename_all = "camelCase")]
 pub struct CourseRevision {
     title: String,
-    content: String,
+    description: String,
```
inyono, line 9: This should still be `content` to not introduce a breaking change.

inyono (changes requested):

src/uuid/model/entity_revision/course_revision.rs
elbotho: ah I see, just change the field it expects from the db then.

inyono: The contract tests catched that btw, yay 🎉

inyono (approved):

PR #103, opened 2021-05-31
Checkout revision: also set `trashed: false`

### LINKED ISSUE #102: When a revision is checked out, set trashed to false
Currently trashed revisions can be checked out but are still trashed afterwards.
(related #100)

Note: This is probably a change from the legacy behaviour. The UI used to only offer the option to "restore" a rejected revision. I simplified this in the frontend to allow a checkout for all revisions.
If this is undesired for some reason, please tell me :)

elbotho:
closes #102 

Note to self: When this is merged we can remove 1-2 lines in the frontend.

### DIFF
--- a/src/uuid/model/entity/mod.rs
+++ b/src/uuid/model/entity/mod.rs
@@ -481,6 +481,17 @@ impl Entity {
 
                 sqlx::query!(
                     r#"
+                        UPDATE uuid
+                            SET trashed = false
+                            WHERE id = ?
+                    "#,
+                    revision_id
+                )
+                .execute(&mut transaction)
+                .await?;
+
+                sqlx::query!(
+                    r#"
                         UPDATE entity
                             SET current_revision_id = ?
                             WHERE id = ?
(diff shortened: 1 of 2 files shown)


src/uuid/model/entity/mod.rs
```diff
@@ -479,6 +479,17 @@ impl Entity {
...
+                sqlx::query!(
+                    r#"
+                        UPDATE uuid
+                            SET trashed = false
+                            WHERE id = ?
+                    "#,
+                    revision_id
+                )
+                .execute(&mut transaction)
+                .await?;
```
kulla, line 491: For a follow up issue: I guess a function `setState()` or `setTrashed()` would be good here...

kulla (approved):

elbotho:
@kulla this is also live in staging right? (seems not to fix the issue unfortunately)

kulla:
> @kulla this is also live in staging right? (seems not to fix the issue unfortunately)

yeah should be live

Issue #107, opened 2021-06-10
Add tests for omiting notifications


kulla:
Follow up of #106 : 

* Omit notifications whose uuid parameter is not supported
* Add tests

PR #108, opened 2021-06-10
Follow up for PR with omiting unsupported notifications

kulla:
Closes https://github.com/serlo/serlo.org-database-layer/issues/107

### DIFF
--- a/src/notification/model.rs
+++ b/src/notification/model.rs
@@ -28,7 +28,7 @@ impl From<sqlx::Error> for NotificationsError {
     }
 }
 
-#[derive(Serialize)]
+#[derive(Serialize, Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct Notification {
     pub id: i32,
@@ -84,8 +84,8 @@ impl Notifications {
                     WHERE n.user_id = ?
                       AND uuid1.discriminator != "attachment"
                       AND uuid1.discriminator != "blogPost"
-                      AND uuid2.discriminator != "attachment"
-                      AND uuid2.discriminator != "blogPost"
+                      AND (uuid2.discriminator IS NULL OR uuid2.discriminator != "attachment")
+                      AND (uuid2.discriminator IS NULL OR uuid2.discriminator != "blogPost")
                       AND (entity1.type_id IS NULL OR entity1.type_id IN (1,2,3,4,5,6,7,8,49,50))
                       AND (entity2.type_id IS NULL OR entity2.type_id IN (1,2,3,4,5,6,7,8,49,50))
                     ORDER BY n.date DESC, n.id DESC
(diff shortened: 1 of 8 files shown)


src/notification/model.rs
```diff
@@ -76,10 +76,17 @@ impl Notifications {
...
+                    LEFT JOIN uuid uuid2 on uuid2.id = event_parameter_uuid.uuid_id
+                    LEFT JOIN entity entity2 on entity2.id = event_parameter_uuid.uuid_id
                     WHERE n.user_id = ?
-                      AND uuid1.discriminator != "attachment"
-                      AND uuid1.discriminator != "blogPost"
+                      AND uuid1.discriminator NOT IN ("attachment", "blogPost")
+                      AND (uuid2.discriminator IS NULL OR
+                        uuid2.discriminator NOT IN ("attachment", "blogPost"))
                       AND (entity1.type_id IS NULL OR entity1.type_id IN (1,2,3,4,5,6,7,8,49,50))
+                      AND (entity2.type_id IS NULL OR entity2.type_id IN (1,2,3,4,5,6,7,8,49,50))
```
inyono, line 89: A bit hard to read & understand (SQL ftw I guess...). Does this check whether UUIDs that occur in the event params also have the right type? I'm not entirely sure whether this works correctly for events with two UUID parameters since I think that you only check the first UUID parameter (because of `LEFT JOIN`).

inyono (changes requested):

src/notification/model.rs
```diff
@@ -486,4 +596,49 @@ mod tests {
...
+
+        sqlx::query!(
+            r#"
+                INSERT INTO user (id, username, email, password, token)
+                VALUES (?, ?, ?, "", ?)
+            "#,
+            new_user_id,
+            random_string(10),
+            random_string(10),
+            random_string(10)
```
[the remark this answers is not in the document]
kulla: Since `username` and `email` is set to be unique in the MySQL having default values would not work when create_test_user() is called twice or when a transactions was saved in the database (In debugging I put a transaction.commit() at the end of the test and checked the changed database values to find bugs in my code). Having a way to produce unique values here would be great (can be deterministic, though). Do you have an idea?

src/notification/model.rs
kulla: > Does this check whether UUIDs that occur in the event params also have the right type?

Yes

>  I'm not entirely sure whether this works correctly for events with two UUID parameters since I think that you only check the first UUID parameter (because of `LEFT JOIN`).

Are there any? I thought there would never be such an event...

However: It is possible that event_parameter_uuid is an unsupported uuid type when event_log.uuid is supported?! I check this...

kulla: Okay, I have found an event type where it is necessary to check the uuid_paramater: create_thread  https://github.com/serlo/serlo.org-database-layer/blob/main/src/event/model/create_thread.rs#L19-L31 We might have a thread on an unsupported entity / uuid.

kulla: Okay, I have added code to deduplicate notifications when they have multiple event_parameter_uuid -> however I haven't seen an event with multiple event_paramater_uuid yet...

inyono (approved):

PR #113, opened 2021-06-24
Add "UserActivityByTypeQuery"

kulla:
For https://github.com/serlo/frontend/issues/935

### DIFF
--- a/src/user/messages.rs
+++ b/src/user/messages.rs
@@ -7,11 +7,27 @@ use crate::database::Connection;
 use crate::message::MessageResponder;
 
 #[derive(Deserialize, Serialize)]
-#[serde(tag = "type", content = "payload")]
-pub enum UserMessage {
+#[serde(tag = "type")]
+pub enum UserQueriesMessage {
     ActiveAuthorsQuery(ActiveAuthorsQuery),
     ActiveReviewersQuery(ActiveReviewersQuery),
+}
+
+#[derive(Deserialize, Serialize)]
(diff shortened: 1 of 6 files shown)


src/user/messages.rs
```diff
@@ -8,18 +8,34 @@ use crate::message::MessageResponder;
...
+pub enum UserQueriesMessage {
     ActiveAuthorsQuery(ActiveAuthorsQuery),
     ActiveReviewersQuery(ActiveReviewersQuery),
 }
 
+#[derive(Deserialize, Serialize)]
+#[serde(tag = "type", content = "payload")]
+pub enum UserMessage {
+    UserActivityByTypeQuery(UserActivityByTypeQuery),
+}
```
inyono, line 20: Why do we need a new namespace for that? So far, all messages in module `Foo` are "exported" together as `FooMesssage`.

inyono (changes requested):
LGTM, apart from the open question regarding the new namespace.

src/user/messages.rs
kulla: `ActiveAuthorsQuery` does not have a payload and thus I cannot store it in an enum with `content=payload`. Otherwise we would need to change api.serlo.org so that it sends `payload = {}` for `ActiveAuthorsQuery`.

inyono: Check. Okay, that makes sense for now.

inyono (approved):

PR #114, opened 2021-06-24
Add helper script to run all check (for example before pushing)

kulla:
Hopefully I will open less PRs with errors in the future :smile: :see_no_evil: -> I want to have this in the codebase to make the live for other developers easier.

### DIFF
--- a/scripts/run_all_checks.sh
+++ b/scripts/run_all_checks.sh
@@ -6,10 +6,15 @@ function init() {
   BOLD=$(tput bold)
   NORMAL=$(tput sgr0)
 
+  WAIT_FOR_MYSQL_TIMEOUT=20
+
   read_arguments "$@"
 
   if ! mysql_is_running; then
+    print_header "Start mysql"
     yarn start
+    log "MySQL need to start, let's wait 30 seconds until it has started..."
+    sleep 30
   fi
 }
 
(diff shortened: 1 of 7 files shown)


elbotho:
> Hopefully I will open less PRs with errors in the future 😄 🙈 

Just as a side note: PRs with errors might feel bad, but are not a real problem imo.
That's why we have the checks running here.
Some companies even encourage devs not to slow down their machines by running checks and builds locally :)

Still like the helpers 👍

scripts/run_all_checks.sh
```diff
@@ -0,0 +1,94 @@
...
+	if [ -n "$(git diff sqlx-data.json)" ]; then
+		error "You need to run sqlx:prepare and commit the changes in sqlx-data.json!"
+	fi
+}
+
+function run_pact_tests {
+	# Run db layer server in background
+	cargo run --quiet &
+	DB_LAYER_SERVER_ID=$!
+	sleep 20
```
inyono, line 60: This should probably use something like https://github.com/serlo/serlo.org-database-layer/blob/main/.github/wait-for-server.sh instead.

inyono (changes requested):

kulla:
@inyono Review suggestions implemented. Ready to be merged

inyono (approved):
LGTM. I added `prettier-plugin-sh` so that the bash scripts get also formatted. The only thing I noticed is that this still requires `yarn start` beforehand. We could possibly add that to the script, too (since `yarn start` won't do much if the server has already been started).

PR #118, opened 2021-07-02
getSubscriptions shall only return supported uuids

kulla:
Closes https://github.com/serlo/api.serlo.org/issues/429


elbotho:
Could you ping me when this is deployed in staging so I can test? 
Thanks!

PR #122, opened 2021-07-05
Fix: A CreateThreadEvent needs to be created in start_thread()

### LINKED ISSUE #119: CreateCommentNotificationEvent breaks notifications query
With `Invalid value received from data source`.

Sentry: https://sentry.io/organizations/serlo/issues/2481072554/?environment=staging&project=5385534&query=is%3Aunresolved

Basically uuid `192412` is expected to be a comment but is a solution.

Discovered with the `notifications` query of user `Mathewally` (unread notifs) but I narrowed it down to:
_Minimal query to run into error_
``` gql
{
  notificationEvent(id:413198){
    ... on CreateCommentNotificationEvent{      
      comment{id}
    }
  }
}
```

My theory is that the comments id and objects id are mixed up somewhere in api or db layer.

kulla:
Closes #119

### DIFF
--- a/src/event/model/create_comment.rs
+++ b/src/event/model/create_comment.rs
@@ -31,19 +31,17 @@ impl TryFrom<&AbstractEvent> for CreateCommentEvent {
 pub struct CreateCommentEventPayload {
     raw_typename: RawEventType,
     actor_id: i32,
-    object_id: i32,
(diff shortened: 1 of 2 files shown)


inyono (approved):
LGTM

Issue #126, opened 2021-07-15
Allow comments on User


elbotho:
Explore if we can easily allow comments on users by falling back on instance `de` here:

https://github.com/serlo/serlo.org-database-layer/blob/0aac3adf1bdf8dd69ba80d2f6a20c8299130e722/src/thread/model.rs#L330

It's for a feature MVP so don't spend too much time on it. 
If it does not work, thats fine for now.

PS: that's my guess that this is why commenting on users is not working, but it might also be deactivated someplace else

PR #127, opened 2021-07-16
Allow creating threads on user (needed for an MVP)

kulla:
Closes https://github.com/serlo/serlo.org-database-layer/issues/126

@elbotho @inyono How is first can make the review :smile:

### DIFF
--- a/src/thread/model.rs
+++ b/src/thread/model.rs
@@ -441,49 +441,56 @@ mod tests {
 
     #[actix_rt::test]
     async fn start_thread() {
-        for object_id in [1565, 1] {
-            let pool = create_database_pool().await.unwrap();
-            let mut transaction = pool.begin().await.unwrap();
+        run_test_start_thread(1565).await;
+    }
 
-            Threads::start_thread(
-                ThreadStartThreadPayload {
-                    title: "title".to_string(),
-                    content: "content-test".to_string(),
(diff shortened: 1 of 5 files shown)


elbotho:
first ;)
looks good, but @inyono should still review this… 
especially the sql.

src/thread/model.rs
```diff
@@ -390,7 +391,7 @@ impl Threads {
             .save(&mut transaction)
             .await?;
 
-        if payload.subscribe {
+        if payload.subscribe && uuid_data.discriminator != "user" {
```
inyono, line 394: Does this make sense? Shouldn't I still get notifications for threads created at user profiles (or wherever those are displayed)?

src/thread/model.rs
```diff
@@ -424,36 +425,49 @@ mod tests {
...
-                object_id: 1565,
-                user_id: 1,
-                subscribe: true,
-                send_email: false,
-            },
-            &mut transaction,
-        )
-        .await
-        .unwrap();
+        for object_id in [1565, 1] {
```
inyono, line 428: Rather have those in two tests since they test different things.

kulla:
@inyono Thx for the feedback. Added both suggestions. Ready for merge.

inyono (approved):

PR #134, opened 2021-08-02
Remaining changes for refactoring EventsQuery

kulla:


src/event/messages.rs
```diff
@@ -67,18 +68,33 @@ impl MessageResponder for EventQuery {
 pub struct EventsQuery {
     after: Option<i32>,
     actor_id: Option<i32>,
+    object_id: Option<i32>,
+    first: i32,
```
elbotho, line 72: tiniest thing but I'd preferre the parameters to be ordered something like `first,after,actor_id,object_id` here and also for the `Events::fetch…` functions.

elbotho (approved):
lgtm, does it work? 😅

inyono (commented):
LGTM

PR #138, opened 2021-08-09
Refactoring of src/subject and src/user

kulla:
This PR aims at making the code less complex (~120 lines of code are removed). The main changes are:

* Remove the implementation of MessageResponder at the lowest level, since it is not really necessary here (there can also be functions directly called). This also allows to have endpoints with and without a payload in one message enum (see src/user)
* Use `sqlx::Error` when there it is sufficient. We can implement more errors when they become necessary.
* Use something like `SubjectsQuery(Option<serde_json::Value>)` for an endpoint without a payload so that it can be accessed by `{ type: "SubjectsQuery" }` and `{ type: "SubjectsQuery", payload: {} }`. This will make our infrastructure more robust since we do not need to take care whether the extra payload is send or not.


src/subject/messages.rs
```diff
@@ -1,54 +1,88 @@
...
 
-use super::model::{Subjects, SubjectsError};
 use crate::database::Connection;
 use crate::message::MessageResponder;
 
 #[derive(Deserialize, Serialize)]
 #[serde(tag = "type", content = "payload")]
 pub enum SubjectsMessage {
-    SubjectsQuery(SubjectsQuery),
+    SubjectsQuery(Option<serde_json::Value>),
```
inyono, line 12: Can't we change this to
```suggestion
    SubjectsQuery,
```

src/subject/messages.rs
```diff
@@ -1,54 +1,88 @@
...
-            Err(e) => {
-                println!("/subjects: {:?}", e);
-                match e {
-                    SubjectsError::DatabaseError { .. } => {
-                        HttpResponse::InternalServerError().finish()
-                    }
-                    SubjectsError::InvalidInstance => HttpResponse::InternalServerError().finish(),
-                }
-            }
+async fn subjects_query(connection: Connection<'_, '_>) -> HttpResponse {
```
inyono, line 25: Mh... I prefer to previous pattern. When using `impl MessageResponder`, it's clearer that this function is `SubjectsQuery`-specific and not intended to be used by other code.

src/subject/messages.rs
```diff
@@ -1,54 +1,88 @@
...
+            HttpResponse::InternalServerError().finish()
         }
     }
 }
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Subjects {
+    pub subjects: Vec<Subject>,
+}
```
inyono, line 45: For consistency, I'd keep that in `model`. This way, we have a clear separation of concerns:
- `messages` defines the accepted messages and how they handle possible errors.
- `model` does the internal stuff, i.e. talk to database etc.

src/subject/messages.rs
```diff
@@ -1,54 +1,88 @@
...
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Subject {
+    instance: String,
+    taxonomy_term_id: i32,
+}
+
+async fn fetch_subjects<'a, E>(executor: E) -> Result<Subjects, sqlx::Error>
```
inyono, line 54: How did you get rid of `SubjectsError::InvalidInstance`?

src/user/messages.rs
```diff
@@ -2,124 +2,87 @@ use actix_web::HttpResponse;
...
-pub enum UserQueriesMessage {
-    ActiveAuthorsQuery(ActiveAuthorsQuery),
-    ActiveReviewersQuery(ActiveReviewersQuery),
-}
-
 #[derive(Deserialize, Serialize)]
 #[serde(tag = "type", content = "payload")]
 pub enum UserMessage {
+    ActiveAuthorsQuery(Option<serde_json::Value>),
+    ActiveReviewersQuery(Option<serde_json::Value>),
```
inyono, line 13: Similar as above.
```suggestion
    ActiveAuthorsQuery,
    ActiveReviewersQuery,
```

inyono (changes requested):
I'm not quite convinced about this refactoring, maybe we should talk about that again:

- `impl MessageResponder` made it quite clear what this function is supposed to do. This might also be relevant if we change up `MessageResponder` later on (e.g. because we don't need the weird connection split-up anymore). 
- The models make a clear cut between the "consumer view" (i.e. messages and HTTP responses) vs. "internal view" (database stuff, actual errors)

src/subject/messages.rs
kulla: We fetch all subjects of all instances and make the filtering later in the API -> so we need no instance filter here any more

src/subject/messages.rs
kulla: With your suggested change `{ type: SubjectsQuery, payload: {} }` would result in `400 Bad Request`. With `SubjectsQuery(Option<serde_json::Value>)` we do not need to take care that we omit the payload argument.

src/subject/messages.rs
kulla: I think we need to talk about this in a meeting. The split can be improved. Before the refactoring the struct for the input and for the output are in different files which does not seem to be the best solution. The quoted struct for example is so specific to the query that it cannot be used anywhere else...

kulla:
@inyono Feedback implemented. Ready for new one :-)

src/message.rs
```diff
@@ -20,6 +23,51 @@ pub trait MessageResponder {
...
+impl From<sqlx::Error> for OperationError {
+    fn from(error: sqlx::Error) -> Self {
+        OperationError::InternalServerError(Box::new(error))
+    }
+}
+
+pub type OperationResult<T> = Result<T, OperationError>;
+
+#[async_trait]
+pub trait Operation {
```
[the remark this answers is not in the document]
kulla: close since we have discussed about this...

kulla:
@inyono Ready to be merged...

inyono (approved):

PR #141, opened 2021-09-04
VocabularyTaxonomyQuery

inyono:
Adds `VocabularyTaxonomyQuery` that outputs the taxonomy for the given instance as a Turtle document (to be processed by SkoHub Vocabs).

## TODOs:
- [x] Remove blog.
- [x] Remove community -> Question for dev meeting: where do we save whether a term is "community"?.
- [x] Probably split up query into taxonomy and curriculum (additional parameter in payload?).
- [x] i18n for `ConceptScheme` title.
- [ ] Sorting? Not sure if that is supported by Turtle / skos.
- [x] Better error handling.
- [x] Message naming,
- [ ] Deploy vocabulary.serlo.org -> own repo.


src/vocabulary/messages.rs
```diff
@@ -0,0 +1,67 @@
...
+                    match Vocabulary::fetch_taxonomy_vocabulary(self.instance.clone(), transaction)
+                        .await
+                    {
+                        Ok(result) => Ok(result),
+                        Err(error) => Err(operation::Error::InternalServerError {
+                            error: Box::new(error),
+                        }),
+                    }
+                }
+            }
```
kulla, line 64:
```suggestion
            Ok(match connection {
                Connection::Pool(pool) => Vocabulary::fetch_taxonomy_vocabulary(self.instance.clone(), pool).await?,
                Connection::Transaction(transaction) => Vocabulary::fetch_taxonomy_vocabulary(self.instance.clone(), transaction).await?
            })
```

kulla: The inner match statement is basically: `....await.map_err(|err| err.into())` which is also done by the `?` operator.

src/vocabulary/model.rs
```diff
@@ -0,0 +1,284 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum TaxonomyType {
+    Root, // Level 0
+
+    Blog, // below Root
+
+    ForumCategory, // below Root or ForumCategory
+    Forum,         // below ForumCategory
```
kulla, line 196: Do we want to report them or shall we filter them?

kulla: (They are filtered out by https://github.com/serlo/serlo.org-database-layer/pull/141/files?diff=split&w=1#diff-90656dae891493653ff080c4b5d9b9ce1fd3fa012e10b95c5eec2c49ab10c7a4R102-R104 right?)

kulla (changes requested):

src/vocabulary/model.rs
inyono: As long as we have such entries in the database (which we do), we need to list them here, otherwise https://github.com/serlo/serlo.org-database-layer/pull/141/files?diff=split&w=1#diff-90656dae891493653ff080c4b5d9b9ce1fd3fa012e10b95c5eec2c49ab10c7a4R100 will fail.

src/vocabulary/model.rs
```diff
@@ -120,6 +143,11 @@ impl Vocabulary {
                     &Literal::<String>::new_lang_unchecked(creator, lang.clone()),
                 )?;
             } else if let Some(parent_id) = term.parent_id {
+                if terms_blocklist.contains(&parent_id) {
+                    terms_blocklist.insert(term.id);
+                    continue;
+                }
```
kulla, line 149: It is guaranteed that the terms come in an order that the parents come first due to the ordering by id, right?

kulla (approved):
Just one thing. IMHO ready to be merged

src/vocabulary/model.rs
inyono: Well, usually? ^^' But true, theoretically it is possible that `child.id < parent.id`...

inyono: I'll think about that edge case.

inyono:
Btw: you can see the current result at https://vocabulary.serlo.org

kulla:
> Btw: you can see the current result at https://vocabulary.serlo.org

Nice! How have you published it without changing other repos?! :thinking:

src/vocabulary/model.rs
inyono: Fixed via https://github.com/serlo/serlo.org-database-layer/pull/141/commits/688c99cabb2ae677d85751670907247deff6f020

PR #145, opened 2021-10-12
UserDeleteBotsMutation

elbotho:
- currently only uses `sqlx::Error` (like other simple mutations)
- no tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,11 @@
 
 # Serlo.org Database Layer
 
+The database layer provides a Restful API in front of the database of Serlo.
+
 ## Setup
 
-You need [Docker](https://docs.docker.com/engine/installation/) and [Rust](https://www.rust-lang.org) installed on your system.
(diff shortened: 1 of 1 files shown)


src/user/model.rs
```diff
@@ -92,6 +92,29 @@ impl User {
...
+    {
+        let mut transaction = executor.begin().await?;
+
+        for user_id in &payload.user_ids {
+            sqlx::query!(
+                r#"DELETE FROM uuid WHERE id = ? AND discriminator = 'user'"#,
+                user_id,
+            )
+            .execute(&mut transaction)
+            .await?;
```
kulla, line 110: This should return a MySqlQueryResult which has a method https://docs.rs/sqlx/0.5.9/sqlx/mysql/struct.MySqlQueryResult.html#method.rows_affected giving the number of rows affected. Can you have a check whether it is `1` and raise an error when it is not the case?

kulla (changes requested):
Should work. I only would like to add tests and another check (e.g. is the id really a user).

Issue #144, opened 2021-10-08
Add EntitiesQuery


kulla:
### Todos

- [x] Filter out the following types in the SQL query (the `type` is stored in the sql table `type`):

```
+-----------------------------------------+
| name                                    |
+-----------------------------------------+
| applet                                  |
| article                                 |
| course                                  |
| text-exercise                           |
| text-exercise-group                     |
| video                                   |
+-----------------------------------------+
```

- [x] Filter out trashed objects (the property `trashed` is stored in the column `uuid.trashed`)
- [x] Filter out entities from a configurable list of taxonomies / the following construct can help: https://github.com/serlo/serlo.org-database-layer/blob/099c7241b3bbc8f36e3ebc1c6e1b0f2bf629e669/src/uuid/model/taxonomy_term.rs#L168-L197
  - [x] Everything under https://de.serlo.org/community/87993/community (`taxonomy_term.id == 87993`) besides everything under https://de.serlo.org/community/106081/neue-f%C3%A4cher-und-themen (`taxonomy_term.id == 106081`)
  - [x] Everything under https://de.serlo.org/informatik/75211/baustelle (`taxonomy_term.id == 75211`)
- [x] Add Parameter "instance" (optional) which filters for the instance (= column `instance.subdomain`)
- [x] Parameter "first" which determine the number of returned values (should be smaller then 10_000, otherwise there shall be an error)
- [x] Add Parameter "lastModified" and return only entites last modified after this date

Note: There a still things to consider before we can finish this task. However we know for sure that we need the above features.

hugotiburtino:
- [X] Add feature tests

kulla:
The output shall be changed from a dict of the form `{ entityIds: number[] }` to a dict `{ entitties: object[] }`. The objects shall the same as in https://nbviewer.org/github/serlo/evaluations/blob/2f19c47fa4ca91cb80ac77f29407edf1f6c3f617/Export%20of%20metadata%20to%20BIRD.ipynb#Fetching-metadata-from-the-DB The code also contains SQL statements and methods how to compute the values.

PR #146, opened 2021-10-12
EntitiesQuery

kulla:
See #144


src/uuid/model/entity/mod.rs
```diff
@@ -404,9 +404,19 @@ impl Entity {
...
+    ///
+    /// See Issue #144
     async fn find_entity_ids<'a, E>(
         after: Option<i32>,
         instance: Option<&String>,
+        first: Option<i32>,
+        last_modified: Option<&String>,
         executor: E,
     ) -> Result<Vec<i32>, sqlx::Error>
     where
```
kulla: Idea: We can add the payload struct as an argument as in https://github.com/serlo/serlo.org-database-layer/pull/145/commits/91b3ef8ca17d42a6db5cb23acaa5c5c5dec96673#diff-53630e1968a2dca78c40a46aa0b0cbd39e68220836815c497abf9120f29269b2R97

src/uuid/model/entity/messages.rs
```diff
@@ -40,12 +40,15 @@ impl MessageResponder for EntityMessage {
...
 mod entities_query {
     use super::*;
+    use crate::operation::Error;
 
     #[derive(Deserialize, Serialize)]
     #[serde(rename_all = "camelCase")]
     pub struct Payload {
         after: Option<i32>,
         instance: Option<String>,
+        first: Option<i32>,
```
kulla, line 50: Okay, I had a look: Can be changed from `Option<i32>` to `i32`

src/uuid/model/entity/mod.rs
```diff
@@ -421,11 +431,16 @@ impl Entity {
                     WHERE entity.id > ?
                     AND (? is NULL or instance.subdomain = ?)
                     AND uuid.trashed = 0
-                    ORDER by entity.id limit 500
+                    AND entity.type_id NOT IN (48, 3, 7, 1, 4, 6)
+                    AND entity.date > ?
```
kulla, line 435: You need to check against `entity_revision.date` like in

```sql
select * from entity join entity_revision on entity.current_revision_id = entity_revision.id where entity_revision.date > Date("2015-10-18");
```

kulla:
@hugotiburtino Before it can be merged the following need to be done:

- [ ] The tests need to be updated to test against the current code
- [ ] We need to check whether it is best practice to use `http` instead of `https` in IRIs (have seen this in other examples)
- [ ] Maybe we want to move the code into an extra module for metadata?!

PR #148, opened 2021-10-19
cargo workspace, minor improvements

inyono:


Cargo.toml
```diff
@@ -1,29 +1,6 @@
...
-sentry-actix = { git = "https://github.com/fourbs-net/sentry-rust", branch = "update-actix" }
-serde = "1.0.0"
-serde_json = "1.0.0"
-sophia = "0.7.0"
-sqlx = { version = "0.5.0", features = ["chrono", "offline", "json", "mysql", "runtime-actix-native-tls"] }
-thiserror = "1.0.0"
-
-[dev-dependencies]
-test-utils = { path="test-utils" }
+members = [
```
inyono, line 3: - Makes sure that we have a single `Cargo.lock` (so no differences in SQLX versions for example).
- Makes sure that we have a common `target` directory, so compile times might be faster in some cases.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
+/server/target
```
inyono, line 2: While cargo workspace has a common target directory in `/target`, we still need the project-specific directories because sqlx adds some files 🤷‍♂️

kulla (approved):
Thx!

PR #155, opened 2021-10-29
Restructure + rename modules in integration tests

kulla:
See https://github.com/serlo/serlo.org-database-layer/blob/restructure-tests/server/tests/user.rs for an example. The test output then looks like:

```
     Running tests/user.rs (target/debug/deps/user-1b051eb3e1ac0173)

running 5 tests
test user_activity_by_type_query::returns_user_activity ... ok
test user_potential_spam_users_query::fails_when_first_parameter_is_too_high ... ok
test user_delete_bots_mutation::deletes_a_user_permanentely ... ok
test user_potential_spam_users_query::returns_user_with_a_description ... ok
test user_potential_spam_users_query::with_after_parameter ... ok
```

I am also fine with storing those integration tests in `src/...`, though.

Shall we rewrite most tests into this scheme?


inyono (approved):
Probably fine and more readable. Maybe we could check whether naming the tests module `tests` is a best-practice in the Rust community? If so, `foo_tests` might be an alternative.

PR #159, opened 2021-11-01
Remove AliasError

kulla:
After looking into the source code I found that `AliasError` can be
directly replaced with with the new type `operation::Error` since this error
type includes all necessary information necessary to handle an alias
query. Only when more fine grained information are necessary we should
include other error types.

### DIFF
--- a/server/src/operation.rs
+++ b/server/src/operation.rs
@@ -18,8 +18,11 @@ pub enum Error {
 
 impl From<sqlx::Error> for Error {
     fn from(error: sqlx::Error) -> Self {
-        Error::InternalServerError {
-            error: Box::new(error),
+        match error {
+            sqlx::Error::RowNotFound => Error::NotFoundError,
+            _ => Error::InternalServerError {
+                error: Box::new(error),
+            },
         }
     }
 }
--- a/server/src/alias/messages.rs
+++ b/server/src/alias/messages.rs
@@ -2,7 +2,7 @@ use actix_web::HttpResponse;
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 
-use super::model::{fetch, fetch_via_transaction, AliasError};
+use super::model::{fetch, fetch_via_transaction};
 use crate::database::Connection;
 use crate::instance::Instance;
 use crate::message::MessageResponder;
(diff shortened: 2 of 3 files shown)


server/src/operation.rs
```diff
@@ -18,8 +18,11 @@ pub enum Error {
 
 impl From<sqlx::Error> for Error {
     fn from(error: sqlx::Error) -> Self {
-        Error::InternalServerError {
-            error: Box::new(error),
+        match error {
+            sqlx::Error::RowNotFound => Error::NotFoundError,
+            _ => Error::InternalServerError {
+                error: Box::new(error),
+            },
```
inyono, line 25: I see two potential problems with this change:
- `sqlx::Error::RowNotFound` doesn't necessarily map to 404. While we still can catch a `RowNotFound` error beforehand and map it to something else, it becomes easier to miss such a potential bug because it's easy to just add a `?`.
- I'd make the `_` explicit (i.e. `Error::InternalServerError`, or did we add additional errors here?), so that devs are required to think about what to do with potential new variants added to `Error` (instead of mapping them to a 500) .

inyono (changes requested):

server/src/operation.rs
kulla: For the first point I'll think about it. Your second point I do not quite understand: It is a match against `sqlx::Error` and it means that other sqlx:Error is mapped to `Error::InternalServerError` (which was already the implementation before). Do you want to have a more fine grained `operation::Error` type?! Do you want that the conversion to the 500 response shall not be the default one?

inyono: I'm thinking more about the potential case when we add additional errors, e.g. `FooError::NotFoundError`. With `_`, you might forget that you want to map `FooError::NotFoundError` to 404 for example. If we make it explicit instead, you have to think about it and decide what you want to do with it.

kulla: @inyono Okay, I have reverted the change. Will need to make the error handling in the database better anyway...

kulla:
@hugotiburtino I guess this PR is in a form that also you can review it.

hugotiburtino (approved):

kulla:
@inyono I have merged this PR after the review of @hugotiburtino I can revert if it is necessary

PR #162, opened 2021-11-30
Temporary branch reflecting the staging branch

kulla:
Includes #146 until it is finished


kulla:
not needed anymore since #146 can be merged

PR #166, opened 2022-01-25
Add UserSetEmailMutation

hugotiburtino:


hugotiburtino:
At the test it would be good to check if the email was indeed updated, but it would demand a lot of work now. Let me put it in my TODO list and move on.

server/tests/user.rs
```diff
@@ -126,3 +126,22 @@ mod user_set_description_mutation {
...
+        let user_id = create_new_test_user(&mut transaction).await.unwrap();
+
+        let mutation_response = Message::new(
+            "UserSetEmailMutation",
+            json!({ "userId": user_id, "email": "user@example.com".to_string() }),
+        )
+        .execute_on(&mut transaction)
+        .await;
+
+        assert_ok(mutation_response, json!({ "success": true })).await;
```
kulla, line 145: Please add a check that the the mail was changed properly (via sqlx query)

kulla (changes requested):

kulla (approved):

PR #169, opened 2022-01-27
Upgrade Node to version 16

hugotiburtino:


hugotiburtino:
Needs resolution of problem with node-gyp (see Pacts / test at Actions.)

kulla:
> Needs resolution of problem with node-gyp (see Pacts / test at Actions.)

In this PR `yarn.lock` got deleted. Maybe not intended... :smile:

hugotiburtino:
> In this PR yarn.lock got deleted. Maybe not intended... 😄
😁 I've chosen the strategy to refresh it entirely with `yarn install`.

Anyways, I reinserted the old `yarn.lock` and it was still not working.
I installed nodegit (that uses node-gyp) as a dependecy, upgraded the lock file, and removed nodegit, and although it works now locally it doesn't in Actions.

There is an old issue about nodegit and node 16. https://github.com/nodegit/nodegit/issues/1840

kulla:
Okay, I see. We need to go through the dependencies anyway and update them... Maybe we can replace `nodegit`?!

hugotiburtino:
> Okay, I see. We need to go through the dependencies anyway and update them... Maybe we can replace nodegit?!

It is not our direct dependency, but a dependency from a dependency. Fortunately, one can find it only at inyono/changelog and Jonas has already released the version 0.4.0 of it, which does not use nodegit.  I've upgraded it and it works now.

kulla (approved):

PR #170, opened 2022-01-31
Add EntityAddRevisionMutation

hugotiburtino:


server/src/subscription/model.rs
```diff
@@ -170,9 +170,25 @@ impl Subscription {
...
+        .0
+        .into_iter()
+        .find(|subscription| subscription.user_id == user_id);
+    Ok(subscription)
+}
+
 #[cfg(test)]
 mod tests {
-    use super::{Subscription, Subscriptions};
+    use super::{fetch_subscription_by_user_and_object, Subscription, Subscriptions};
```
inyono, line 191: Shouldn't it be possible to make `mod tests` public and import from that, instead? The disadvantage from your proposed change is that `fetch_subscription_by_user_and_object` is part of the production code.

server/src/uuid/model/entity/mod.rs
```diff
@@ -405,6 +407,128 @@ impl Entity {
...
+        executor: E,
+    ) -> Result<Uuid, EntityAddRevisionError>
+    where
+        E: Executor<'a>,
+    {
+        let mut transaction = executor.begin().await?;
+
+        if let Err(..) = Entity::fetch_via_transaction(payload.entity_id, &mut transaction).await {
+            return Err(EntityAddRevisionError::EntityNotFound);
+        }
```
inyono, line 422: That looks dangerous. The `UuidError` returned via `fetch_via_transaction` covers a couple of cases that don't necessarily mean that the entity does not exists, see https://github.com/serlo/serlo.org-database-layer/blob/main/server/src/uuid/model/uuid.rs#L42.

We should be specific here and correctly pass on other errors like `DatabaseError` so that the API does not wrongly tell the user that the entity does not exist.

inyono (changes requested):

server/src/uuid/model/entity/mod.rs
hugotiburtino: Fair enough.
I've mapped to two possible errors in the 306f7d9, altough `DatabaseError` is maybe to broad (that is, by the way, the error that comes in test, namely 'Cannot add or update a child row: a foreign key constraint fail').
I'm not sure anymore if we need an specific `EntityAddRevisionError::EntityNotFound`, or if we just make a `Entity::fetch_via_transaction` and throw the error up.

server/src/subscription/model.rs
hugotiburtino: Good point. 
Since I've used it in another test, it is better to do as you say. 
Done in cae0f5b

hugotiburtino:
@inyono before I go on writing the remaining tests I would like to get your feedback again, in order to maybe avoid having to refactor the tests.

server/src/uuid/model/entity/mod.rs
inyono: It should be enough to handle `UuidError::NotFound` since `Entity::fetch_via_transaction` should already catch the relevant database errors :)

server/src/uuid/model/entity/mod.rs
```diff
@@ -484,18 +485,81 @@ impl Entity {
...
+                HashMap::from([("content".to_string(), payload.input.content.unwrap())])
+            }
+            EntityRevisionType::Video => HashMap::from([
+                ("content".to_string(), payload.input.content.unwrap()),
+                (
+                    "description".to_string(),
+                    payload.input.description.unwrap(),
+                ),
+                ("title".to_string(), payload.input.title.unwrap()),
+            ]),
```
inyono, line 552: I'd move that part into api.serlo.org and just pass an object `fields` to the database layer. If you want, you could write some short validation code that checks whether the given `fields` contain all relevant data required for the specific entity. But even that you could skip, IMHO, since api.serlo.org has the validation via GraphQL.

server/src/uuid/model/entity/mod.rs
hugotiburtino: alright, done in 463ad8a.
Thanks for your feedback

server/src/uuid/model/entity/mod.rs
hugotiburtino: copy that.
Done in 90bb326 and in https://github.com/serlo/api.serlo.org/pull/525/commits/699e6e7b06fe0664020d447e30af3515b101c28f.
Please check if that implementation is what you suggested

server/src/uuid/model/entity/messages.rs
```diff
@@ -33,6 +37,68 @@ impl MessageResponder for EntityMessage {
...
+                Entity::add_revision(payload, transaction).await
+            }
+        };
+
+        match entity_revision {
+            Ok(_) => HttpResponse::Ok()
+                .content_type("application/json; charset=utf-8")
+                .json(EntityRevisionData {
+                    success: true,
+                    reason: None,
```
hugotiburtino, line 77: @inyono What do you think of returning the `id` of the revision? Could that be somehow useful? At any rate, would make testing a little bit easier

server/src/uuid/model/entity/mod.rs
hugotiburtino: At the end of the day, since 90bb326 we are not using EntityRevisionType. Should I remove it from payload?

hugotiburtino: I'm seeing now the disadvantage that the way I've implemented some fields are send in the mutation with a name, but retrieved with another (for example, CourseRevision sends a 'description' but its content it is fetched in a UuidQuery as 'content'; VideoRevision is sends a 'content', but in the UuidQuery it is a 'url').
I can try to improve that in a follow up PR.

inyono (approved):
LGTM

server/src/uuid/model/entity/messages.rs
hugotiburtino: Good. I'll make it in another PR then.

PR #171, opened 2022-01-31
Fix for pacts

kulla:
Include fixes for breaking pact tests:

* LicenseQuery shall return 404 when license cannot be found
* UserSetEmailQuery shall also return username

(There is still something broken, however it should be with the code in api / I'll have a look tomorrow)

### DIFF
--- a/server/src/user/model.rs
+++ b/server/src/user/model.rs
@@ -187,17 +187,24 @@ impl User {
     pub async fn set_email<'a, E>(
         payload: &user_set_email_mutation::Payload,
         executor: E,
-    ) -> Result<(), sqlx::Error>
+    ) -> Result<String, sqlx::Error>
     where
         E: Executor<'a>,
     {
+        let mut transaction = executor.begin().await?;
+
+        let username = sqlx::query!("select username from user where id = ?", payload.user_id)
+            .fetch_one(&mut transaction)
+            .await?
+            .username;
         sqlx::query!(
             "update user set email = ? where id = ?",
             payload.email,
             payload.user_id
         )
-        .execute(executor)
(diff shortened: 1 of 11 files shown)


server/src/user/model.rs
```diff
@@ -187,17 +187,24 @@ impl User {
...
         payload: &user_set_email_mutation::Payload,
         executor: E,
-    ) -> Result<(), sqlx::Error>
+    ) -> Result<String, sqlx::Error>
     where
         E: Executor<'a>,
     {
+        let mut transaction = executor.begin().await?;
+
+        let username = sqlx::query!("select username from user where id = ?", payload.user_id)
```
hugotiburtino, line 196: That does the job, that shouldn't be anyway complicated. My tendency would be to use a method of the uuid::model::user model.

hugotiburtino:
The test for user_set_email is also failing.

server/src/user/model.rs
kulla: > That does the job, that shouldn't be anyway complicated. My tendency would be to use a fetch method of the uuid::model::user model. Do you think it is easy to use a fetch method instead?

can you try to implement it?

kulla:
@hugotiburtino Ready for merge

hugotiburtino (approved):

server/src/user/model.rs
hugotiburtino: > can you try to implement it?
I'll try in another moment, putting in my TODO list.

hugotiburtino: I've managed to do it locally, but it's not worthy, too much code for too little. Let's leave as it is.

PR #173, opened 2022-02-24
Add PageAddRevisionMutation

hugotiburtino:
- feat(entity): add basic implementation of EntityAddRevisionMutation
- test(entity-add-revision): test creation of revision
- feat(entity-add-revision): validate if entity exists
- feat(entity-add-revision): handle needsReview parameter
- refactor(subscription): make fetch_subscription_by_user_and_object public
- feat(entity-add-revision): handle subscribeThis and subscribeThisByEmail
- feat(entity-add-revision): subscribe also to the entity
- refactor(entity-add-revision): give empty reason to checkout
- refactor(entity-add-revision): remove subscription of revision, that
- feat(event): implement CreateEntityRevisionEvent and use it in
- refactor(entity-add-revision): fill entity_revision_field in just one
- refactor(entity-revision-event): ignore entity revision id
- refactor(entity-add-revision): move entity revision adding logic to
- refactor(entity-add-revision): small fix after refactor 2c4d22d
- refactor(subscriptions): put fetch_subscription_by_user_and_object
- test(entity-add-revision): prefer to fetch revision using ORDER BY
- refactor(entity-add-revision): map errors for checking entity exists
- refactor(entity-add-revision): fit EntityAddRevisionPayload to communicate with api
- Fix by linter
- refactor(entity-add-revision): remove unnecessary decorators
- test(entity-add-revision): try to fix CI-CD 'Deadlock found' error
- feat(entity-add-revision): fill fields according to each revision type
- refactor(entity-add-revision): do not handle DatabaseError and remove
- refactor(entity-add-revision): implement fields param in payload
- test(entity-add-revision): start writing integration tests
- refactor(entity-add-revision): be sure to store field as snake case
- test(entity-add-revision): complete integration tests
- feat(page): implement addRevision mutation
- refactor(page-add-revision): checkout revision after adding it

### DIFF
--- a/server/src/uuid/model/page/mod.rs
+++ b/server/src/uuid/model/page/mod.rs
@@ -119,6 +119,91 @@ impl UuidFetcher for Page {
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
+pub struct PageAddRevisionPayload {
+    pub content: String,
+    pub title: String,
+    pub page_id: i32,
+    pub user_id: i32,
+}
+
+#[derive(Error, Debug)]
+pub enum PageAddRevisionError {
+    #[error("Revision could not be added because of a database error: {inner:?}.")]
+    DatabaseError { inner: sqlx::Error },
+    #[error("Revision could not be added because of an UUID error: {inner:?}.")]
+    UuidError { inner: UuidError },
+    #[error("Revision could not be added because page was not found.")]
+    PageNotFound,
+}
+impl From<sqlx::Error> for PageAddRevisionError {
+    fn from(inner: sqlx::Error) -> Self {
+        Self::DatabaseError { inner }
+    }
+}
+
+impl From<UuidError> for PageAddRevisionError {
+    fn from(error: UuidError) -> Self {
+        match error {
+            UuidError::DatabaseError { inner } => inner.into(),
+            inner => Self::UuidError { inner },
+        }
+    }
+}
+
+impl Page {
+    pub async fn add_revision<'a, E>(
+        payload: PageAddRevisionPayload,
+        executor: E,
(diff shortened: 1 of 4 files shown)


server/src/uuid/model/page/mod.rs
```diff
@@ -117,6 +117,113 @@ impl UuidFetcher for Page {
...
+impl Page {
+    pub async fn add_revision<'a, E>(
+        payload: PageAddRevisionPayload,
+        executor: E,
+    ) -> Result<Uuid, PageAddRevisionError>
+    where
+        E: Executor<'a>,
+    {
+        let mut transaction = executor.begin().await?;
+
```
hugotiburtino, line 174: Should I add any checks if content and title are empty?

inyono: Good point. I'd at least do the validation in the API. Might also be a topic for the dev meeting on how we want to handle validation in general. IMHO:

- API is the consumer-facing service and should handle most of the user-input validation.
- Database layer may assume that API has already done the validation.

I'm open for other suggestions, though. As long as we stay mostly consistent with our approach, I'm fine with it.

server/src/uuid/model/page/mod.rs
```diff
@@ -117,6 +117,113 @@ impl UuidFetcher for Page {
...
+            "#,
+            page_revision_id,
+            payload.user_id,
+            payload.page_id,
+            payload.title,
+            payload.content
+        )
+        .execute(&mut transaction)
+        .await?;
+
```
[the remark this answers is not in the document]
hugotiburtino: Sure, thanks for pointing that out. Done in 7c38ad4

inyono (approved):
LGTM

PR #181, opened 2022-03-22
EntityCreate: handle taxonomyTermId

hugotiburtino:


server/src/uuid/model/entity/mod.rs
```diff
@@ -612,6 +614,27 @@ impl Entity {
...
+                revision_type: EntityRevisionType::from(payload.entity_type.clone()),
+                user_id: payload.user_id,
+            },
+            &mut transaction,
+        )
+        .await?;
+
+        CreateEntityEventPayload::new(entity_id, payload.user_id, instance_id)
+            .save(&mut transaction)
+            .await?;
```
kulla, line 636: Is there a test for this?

kulla: shouldn't this be before the `Entity::addRevision()`? Otherwise we would have a `AddRevisionEvent` before a `CreateEntityEvent`...

server/src/uuid/model/entity/mod.rs
```diff
@@ -639,41 +662,46 @@ impl Entity {
             )
             .save(&mut transaction)
             .await?;
-        }
+        } else {
+            let taxonomy_term_id = payload
+                .input
+                .taxonomy_term_id
+                .ok_or(EntityCreateError::TaxonomyTermIdError)?;
```
kulla, line 669: Use `unwrap_or` here...

kulla (changes requested):

server/src/uuid/model/entity/mod.rs
hugotiburtino:
```console
   --> server/src/uuid/model/entity/mod.rs:669:28
    |
669 |                 .unwrap_or(EntityCreateError::TaxonomyTermIdError)?;
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found enum `EntityCreateError`
```
same with `unwrap_or_else` by the way

server/src/uuid/model/entity/mod.rs
hugotiburtino: good catch
events order fixed in 58297fe
integration test in ed34a2d

server/src/uuid/model/entity/mod.rs
```diff
@@ -639,41 +662,46 @@ impl Entity {
...
+                taxonomy_term_id,
+                current_last_position + 1
+            )
+            .execute(&mut transaction)
+            .await?;
 
-        CreateEntityEventPayload::new(entity_id, payload.user_id, instance_id)
+            CreateTaxonomyLinkEventPayload::new(
+                entity_id,
+                payload.input.taxonomy_term_id.unwrap(),
```
kulla, line 698:
```suggestion
                taxonomy_term_id,
```

server/src/event/model/create_taxonomy_link.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::database::Executor;
+use serde::Serialize;
+
+use super::{AbstractEvent, Event, EventError, EventPayload, RawEventType};
+
+#[derive(Debug, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CreateTaxonomyLinkEvent {
+    entity_id: i32,
+    entity_revision_id: i32,
```
kulla, line 13:
```suggestion
    taxonomy_id: i32,
```

?

kulla (changes requested):

server/src/uuid/model/entity/mod.rs
hugotiburtino: done in [bc2b8c1](https://github.com/serlo/serlo.org-database-layer/pull/181/commits/bc2b8c14b8bd1212269e5c245a807cac0e9c08af)

server/src/event/model/create_taxonomy_link.rs
hugotiburtino: I ask myself too. And there are a lot of mistakes in the file.
Corrected in [bc2b8c1](https://github.com/serlo/serlo.org-database-layer/pull/181/commits/bc2b8c14b8bd1212269e5c245a807cac0e9c08af)

hugotiburtino:
Quite strange this never ending pipeline: https://github.com/serlo/serlo.org-database-layer/runs/5727707102?check_suite_focus=true . It took one hour before and I had to stop it by hand.

server/src/event/model/create_taxonomy_link.rs
hugotiburtino: But I'm not sure about the following asymmetry:
In Create Entity Link Event (https://github.com/serlo/serlo.org-database-layer/blob/entity-create-patch/server/tests/entity.rs#L962-L967) the bearer of the `object_id` is the child.

But in the Create Taxonomy Link Event (https://github.com/serlo/serlo.org-database-layer/blob/entity-create-patch/server/tests/entity.rs#L902-L907) the `object_id` refers to the "parent".

I'm basing on the following result of the DB:
```sql
mysql> select event_log.uuid_id as 'uuid of taxonomy parent', event.name, event_parameter_uuid.uuid_id as 'uuid of entity child' f
rom event_log join event_parameter on log_id = event_log.id join event_parameter_uuid on event_parameter.id = event_parameter_id j
oin event on event.id = event_id  where event_log.id = 299;
+-------------------------+-------------------------+----------------------+
| uuid of taxonomy parent | name                    | uuid of entity child |
+-------------------------+-------------------------+----------------------+
|                       7 | taxonomy/term/associate |                 1495 |
+-------------------------+-------------------------+----------------------+

```

hugotiburtino: clared

kulla (approved):

PR #182, opened 2022-03-24
Add TaxonomyTermSetNameAndDescriptionMutation

hugotiburtino:


server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -236,3 +242,120 @@ impl TaxonomyTerm {
...
+        }
+    }
+}
+
+impl From<TaxonomyTermSetNameAndDescriptionError> for operation::Error {
+    fn from(error: TaxonomyTermSetNameAndDescriptionError) -> Self {
+        match error {
+            TaxonomyTermSetNameAndDescriptionError::DatabaseError {
+                inner: sqlx::Error::RowNotFound,
+            } => operation::Error::NotFoundError,
```
kulla, line 278: Should be a UserInputError: "term with term_id does not exist"

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -236,3 +242,120 @@ impl TaxonomyTerm {
...
+            }
+        }
+    }
+}
+
+impl TaxonomyTerm {
+    pub async fn set_name_and_description<'a, E>(
+        payload: &taxonomy_term_set_name_and_description_mutation::Payload,
+        executor: E,
+    ) -> Result<bool, TaxonomyTermSetNameAndDescriptionError>
```
kulla, line 293:
```suggestion
    ) -> Result<(), operation::Error>
```

Then you will need less code...

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -236,3 +242,120 @@ impl TaxonomyTerm {
...
+
+        let term_id = sqlx::query!(
+            r#"
+                SELECT term_id
+                    FROM term_taxonomy
+                    WHERE id = ?
+            "#,
+            payload.id
+        )
+        .fetch_one(&mut transaction)
```
kulla, line 307: Idea: fetch_optional and map null to user input error

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -236,3 +242,120 @@ impl TaxonomyTerm {
...
+            term_id,
+        )
+        .execute(&mut transaction)
+        .await?;
+
+        let mut description = "".to_string();
+
+        if payload.description.is_some() {
+            description = payload.description.clone().unwrap();
+        }
```
kulla, line 327: `description = payload.unwrap_or("".to_string())

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -236,3 +242,120 @@ impl TaxonomyTerm {
...
+    pub async fn set_name_and_description<'a, E>(
+        payload: &taxonomy_term_set_name_and_description_mutation::Payload,
+        executor: E,
+    ) -> Result<bool, TaxonomyTermSetNameAndDescriptionError>
+    where
+        E: Executor<'a>,
+    {
+        let mut transaction = executor.begin().await?;
+
+        let term_id = sqlx::query!(
```
kulla, line 299: here you can also fetch instance_id -> one less query

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -236,3 +242,120 @@ impl TaxonomyTerm {
...
+pub enum TaxonomyTermSetNameAndDescriptionError {
+    #[error(
+        "Taxonomy Term name and description cannot be set because of a database error: {inner:?}."
+    )]
+    DatabaseError { inner: sqlx::Error },
+    #[error(
+        "Taxonomy Term name and description cannot be set because of an event error: {inner:?}."
+    )]
+    EventError { inner: EventError },
+}
```
kulla, line 256:
```suggestion
```

You should be able to only use `operation::Error`

server/tests/taxonomy_term.rs
```diff
@@ -0,0 +1,31 @@
...
+
+        let query_response = Message::new("UuidQuery", json!({"id": 7}))
+            .execute_on(&mut transaction)
+            .await;
+
+        assert_ok_with(query_response, |result| {
+            assert_eq!(result["name"], "a name");
+            assert_eq!(result["description"], "a description");
+        })
+        .await;
```
kulla, line 29: test for the new event is missing...

kulla (changes requested):

server/tests/taxonomy_term.rs
hugotiburtino: done in e413673

An aside: when testing locally we see this result
```
user@serlo:~ yarn fetch EventsQuery '{ "first": 5, "objectId": 7 }'
...
    {
      "__typename": "SetTaxonomyTermNotificationEvent",
      "id": 86598,
      "instance": "de",
      "date": "2022-03-28T12:12:19+02:00",
      "actorId": 1,
      "objectId": 7,
      "taxonomyTermId": 7
    },
    {
      "__typename": "RemoveTaxonomyLinkNotificationEvent",
      "id": 76642,
      "instance": "de",
      "date": "2014-09-16T09:26:49+02:00",
      "actorId": 27687,
      "objectId": 7,
      "parentId": 7,
      "childId": 18758
    }

```
 Observe objectId 7 and parentId 7 (as well as objectId 7 and taxonomyTermId 7). Should it be so?

server/src/uuid/model/taxonomy_term/mod.rs
hugotiburtino: done in 3f20e00

server/src/uuid/model/taxonomy_term/mod.rs
hugotiburtino: done in 5ae0028

server/src/uuid/model/taxonomy_term/mod.rs
hugotiburtino: instead of `payload.description.clone().unwrap_or("".to_string())`, clippy suggested to use `unwrap_or_else`. 
Done in 5ae0028

server/src/uuid/model/taxonomy_term/mod.rs
hugotiburtino: good idea, unfortunately I couldn't do that in a clean way. Since in the api we would already do this check, we may leave it as it is. Otherwise, we could take a look together.

server/src/uuid/model/taxonomy_term/mod.rs
hugotiburtino: done in 3f20e00

server/src/uuid/model/taxonomy_term/mod.rs
hugotiburtino: alright, done in 3f20e00

kulla (approved):
Do other suggestions in a separate PR

PR #184, opened 2022-03-29
db: add fake user with id 4

inyono:


__tests-pacts__/pacts.js
```diff
@@ -41,6 +41,7 @@ test('Pacts', async () => {
     provider: 'serlo.org-database-layer',
     providerVersion,
     providerBaseUrl: 'http://localhost:8080',
+    consumerVersionTag: '0.29.1-b68d211',
```
inyono, line 44: ping @kulla @hugotiburtino:

If `consumerVersionTag` is set, the contract is tested against that particular version. So this can be used if api.serlo.org is already "ahead" and we haven't implemented that feature in serlo.org-database-layer. See https://pact.serlo.org/matrix/provider/serlo.org-database-layer/consumer/api.serlo.org for a list of possible tags.

PR #189, opened 2022-03-31
Remove EntityAddRevisionError and EntityCreateError

kulla:


server/src/uuid/model/entity/messages.rs
```diff
@@ -84,9 +84,9 @@ pub mod entity_add_revision_mutation {
 
         async fn execute(&self, connection: Connection<'_, '_>) -> operation::Result<Self::Output> {
             let entity_revision = match connection {
-                Connection::Pool(pool) => Entity::add_revision(self, pool).await.unwrap(),
+                Connection::Pool(pool) => Entity::add_revision(self, pool).await?,
                 Connection::Transaction(transaction) => {
-                    Entity::add_revision(self, transaction).await.unwrap()
+                    Entity::add_revision(self, transaction).await?
```
kulla, line 89: BTW: Here is the problem with the pact tests. There is some error and due to this error and `unwrap()` the local server just stops. -> As I said before: Never use `unwrap()` in the production code :smile:

hugotiburtino (approved):

PR #191, opened 2022-04-03
Update README.md

MoeHome:


README.md
```diff
@@ -32,6 +32,10 @@ Run `yarn` to install all necessary node dependencies (needed for development an
 
 For the command `yarn fetch` the tool [`jq`](https://stedolan.github.io/jq/) needs to be installed.
 
+### Install gcc and gcc-multilib
+
+On Ubuntu you might also need [`gcc`](https://gcc.gnu.org/) and gcc-multilib.
```
hugotiburtino, line 37: I'm not sure if we need it.
I'm using Ubuntu in a VM, gcc-multilib is not installed and it works though.
Besides that, we have devs using Mac. 
What led you to install them? What problem you were trying to solve?

kulla: @hugotiburtino Do you have `gcc-multilib` installed on your VM?

hugotiburtino: >  Do you have gcc-multilib installed on your VM?
No I don't

MoeHome: I'm not sure if it was cargo run or cargo test which failed with <pre><font color="#EF2929"><b>error</b></font><b>: linking with `cc` failed: exit status: 1</b></pre> After installing both it worked. But now I do think that the error has to do with something else because it kind of occurs randomly.

README.md
```diff
@@ -32,6 +32,10 @@ Run `yarn` to install all necessary node dependencies (needed for development an
 
 For the command `yarn fetch` the tool [`jq`](https://stedolan.github.io/jq/) needs to be installed.
 
+### Install gcc and gcc-multilib
```
kulla, line 35:
```suggestion
### Install gcc
```

kulla (changes requested):

kulla (approved):

PR #193, opened 2022-04-04
Add_UserDeleteRegularUsersMutation

MoeHome:


server/src/user/model.rs
```diff
@@ -127,6 +128,139 @@ impl User {
...
+            )
+            .execute(&mut transaction)
+            .await?;
+
+        sqlx::query!(
+                r#"delete from uuid where id = ? and discriminator = 'user'"#,
+                payload.id,
+            )
+            .execute(&mut transaction)
+            .await?;
```
hugotiburtino, line 257: I suppose you want to discuss this part. Indeed, not that beautiful.
A workaround would be to use one for-loop for the updates and another for the deletes.
Something like
```
for table in ["subscription", "uuid" /*etc*/"]{ // not sure if it is going to work
sqlx::query!(
                r#"delete from ? where id = ? and discriminator = 'user'"#,
                table,
                payload.id
```

server/src/user/model.rs
```diff
@@ -127,6 +128,139 @@ impl User {
...
+
+        transaction.commit().await?;
+
+        Ok(Ok(()))
```
hugotiburtino, line 261:
```suggestion
        Ok(())
```

hugotiburtino (changes requested):

server/tests/user.rs
```diff
@@ -46,6 +46,179 @@ mod user_delete_bots_mutation {
...
+            deleted_got_activity,
+            json!({ "edits": edits, "reviews": reviews, "comments": comments, "taxonomy": taxonomy,}),
+        )
+            .await;
+
+        let check_ad = sqlx::query!(r#"select author_id from ad where id = 1"#)
+            .fetch_optional(&mut transaction)
+            .await;
+
+        assert_eq!(check_ad.unwrap().unwrap().author_id, deleted_user_id);
```
hugotiburtino, line 93:
```suggestion
            .fetch_one(&mut transaction)
            .await;

        assert_eq!(check_ad.unwrap().author_id, deleted_user_id);
```
Same fot the other lines: (96, 99), (102,105) and so on.

hugotiburtino (changes requested):

server/src/user/model.rs
hugotiburtino: > The problem with the for-loop is that the ids are called different in the different tables (id, user_id, author_id,...)
Good point.
In that case, it is good as it is.

hugotiburtino:
Thanks a lot. The implementation was harder than I initially thought. Well done!

server/src/user/model.rs
```diff
@@ -127,6 +128,116 @@ impl User {
...
+
+        let result = sqlx::query!(r#"select * from user where id = ?"#, payload.id,)
+            .fetch_optional(&mut transaction)
+            .await?;
+
+        if result.is_none() {
+            return Err(operation::Error::BadRequest {
+                reason: "The requested User does not exist.".to_string(),
+            });
+        }
```
kulla, line 155:
```suggestion
        sqlx::query!(r#"select id from user where id = ?"#, payload.id,)
            .fetch_optional(&mut transaction)
            .await?
            .ok_or(|| operation::Error::BadRequest {
                reason: "The requested User does not exist.".to_string(),
            })?;
```

server/tests/user.rs
```diff
@@ -46,6 +46,179 @@ mod user_delete_bots_mutation {
...
+            .execute_on(&mut transaction)
+            .await;
+
+        assert_bad_request(response, "The requested User does not exist.").await;
+    }
+
+    #[actix_rt::test]
+    async fn fails_when_trying_to_delete_deleted() {
+        let mut transaction = begin_transaction().await;
+        let deleted_user_id: i64 = 4;
```
kulla, line 209:
```suggestion
        let deleted_user_id: i32 = 4;
```

kulla (changes requested):

server/src/user/model.rs
MoeHome: Das funktioniert nicht wegen

> error[E0277]: `?` couldn't convert the error to `operation::Error

> `the trait `std::convert::From<[closure@server/src/user/model.rs:150:20: 152:15]>` is not implemented for `operation::Error`

hugotiburtino (approved):

PR #200, opened 2022-04-07
Add TaxonomyTermMoveMutation

hugotiburtino:


server/src/event/model/set_taxonomy_parent.rs
```diff
@@ -28,3 +31,61 @@ impl TryFrom<&AbstractEvent> for SetTaxonomyParentEvent {
         })
     }
 }
+
+pub struct SetTaxonomyParentEventPayload {
+    raw_typename: RawEventType,
+    child_id: i32,
+    previous_parent_id: Option<i64>,
```
kulla, line 38:
```suggestion
    previous_parent_id: i32,
```

We would create a loop when we move root -> should result in `BadRequest`

server/src/event/model/set_taxonomy_parent.rs
```diff
@@ -28,3 +31,61 @@ impl TryFrom<&AbstractEvent> for SetTaxonomyParentEvent {
...
+            self.actor_id,
+            self.child_id,
+            self.instance_id,
+            HashMap::new(),
+            uuid_parameters,
+        )
+        .save(&mut transaction)
+        .await?;
+
+        transaction.commit().await?;
```
kulla, line 87:
```suggestion
        let mut uuid_parameters: HashMap<String, i32> = HashMap::new();

        if let Some(previous_parent_id) = self.previous_parent_id {
            uuid_parameters.insert("from".to_string(), previous_parent_id as i32);
        }

        uuid_parameters.insert("to".to_string(), self.parent_id);

        let event = EventPayload::new(
            self.raw_typename.clone(),
            self.actor_id,
            self.child_id,
            self.instance_id,
            HashMap::new(),
            uuid_parameters,
        )
        .save(executor)
        .await?;
```

Can you check in the other events that they are the same?

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -300,4 +298,80 @@ impl TaxonomyTerm {
...
+                    reason: format!(
+                        "Child cannot have same id {} as destination",
+                        payload.destination
+                    ),
+                });
+            };
+
+            let child = sqlx::query!(
+                r#"
+                    SELECT instance_id, parent_id AS previous_parent_id
```
kulla, line 336:
```suggestion
                    SELECT term.instance_id, term_taxonomy.parent_id AS previous_parent_id
```

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -300,4 +298,80 @@ impl TaxonomyTerm {
...
+                    WHERE term_taxonomy.id = ?
+                "#,
+                child_id
+            )
+            .fetch_optional(&mut transaction)
+            .await?
+            .ok_or(operation::Error::BadRequest {
+                reason: format!("Taxonomy term with id {} does not exist", child_id),
+            })?;
+
```
kulla, line 349: Add check: `destination != previous_parent_id`

server/tests/taxonomy_term.rs
```diff
@@ -69,3 +69,112 @@ mod set_name_and_description_mutation {
...
+            json!({ "childrenIds": [1288], "destination": 1288, "userId": 1 }),
+        )
+        .execute_on(&mut transaction)
+        .await;
+
+        assert_bad_request(response, "Child cannot have same id 1288 as destination").await;
+    }
+
+    #[actix_rt::test]
+    async fn fails_when_parent_or_child_does_not_exist() {
```
kulla, line 145: make two tests here

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -300,4 +298,80 @@ impl TaxonomyTerm {
 
         Ok(())
     }
+
+    pub async fn move_to_new_parent<'a, E>(
```
kulla, line 302: Add check: instance of destination = instance of moved child.

kulla (changes requested):

server/src/event/model/set_taxonomy_parent.rs
hugotiburtino: I'm not sure what should be checked here. I suppose it i the use of `let mut uuid_parameters: HashMap<String, i32> = HashMap::new();` and so on. I've changed. Please take a look later.

kulla (approved):

PR #207, opened 2022-04-12
Add TaxonomyCreateMutation

hugotiburtino:


server/src/event/model/create_taxonomy_term.rs
```diff
@@ -0,0 +1,56 @@
...
+            self.actor_id,
+            self.taxonomy_term_id,
+            self.instance_id,
+            HashMap::new(),
+            HashMap::new(),
+        )
+        .save(executor)
+        .await?;
+
+        Ok(event)
```
kulla, line 54:
```suggestion
        EventPayload::new(
            self.raw_typename.clone(),
            self.actor_id,
            self.taxonomy_term_id,
            self.instance_id,
            HashMap::new(),
            HashMap::new(),
        )
        .save(executor)
        .await
```

server/src/uuid/model/taxonomy_term/messages.rs
```diff
@@ -103,3 +109,35 @@ pub mod taxonomy_term_move_mutation {
         }
     }
 }
+
+pub mod taxonomy_term_create_mutation {
+    use super::*;
+    use crate::instance::Instance;
+    use crate::uuid::Uuid;
+    use crate::vocabulary::model::TaxonomyType;
```
kulla, line 117: shouldn't this be moved to this modul (like in `model.rs`)

server/src/uuid/model/taxonomy_term/messages.rs
```diff
@@ -103,3 +109,35 @@ pub mod taxonomy_term_move_mutation {
         }
     }
 }
+
+pub mod taxonomy_term_create_mutation {
+    use super::*;
+    use crate::instance::Instance;
+    use crate::uuid::Uuid;
```
kulla, line 116: I would move this to the top of the file.

server/src/uuid/model/taxonomy_term/messages.rs
```diff
@@ -103,3 +109,35 @@ pub mod taxonomy_term_move_mutation {
...
+    use crate::vocabulary::model::TaxonomyType;
+
+    #[derive(Deserialize, Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Payload {
+        pub user_id: i32,
+        pub taxonomy_type: TaxonomyType,
+        pub parent_id: Option<i32>,
```
kulla, line 124:
```suggestion
        pub parent_id: i32,
```

let's not create another root :smile: (this should only be done by db migrations...)

server/src/uuid/model/taxonomy_term/messages.rs
```diff
@@ -103,3 +109,35 @@ pub mod taxonomy_term_move_mutation {
...
+        pub instance: Instance,
```
kulla, line 125: actually not needed since we can get the instance by the parent_id... (or we need a check that instance of parent_id = instance). -> I would delete this parameter.

server/tests/taxonomy_term.rs
```diff
@@ -219,3 +219,63 @@ mod move_mutation {
...
 }
+
+#[cfg(test)]
+mod create_mutation {
+    use assert_json_diff::assert_json_include;
+    use test_utils::*;
+
+    #[actix_rt::test]
+    async fn creates_new_taxonomy_term() {
+        // TODO: maybe iterate over taxonomy types to test all?
```
kulla, line 230: that would be great!

server/tests/taxonomy_term.rs
```diff
@@ -219,3 +219,63 @@ mod move_mutation {
...
+        let mut transaction = begin_transaction().await;
+
+        let mutation_response = Message::new(
+            "TaxonomyTermCreateMutation",
+            json!({
+                "parentId": 1394,
+                "name": "a name",
+                "description": "a description",
```
kulla, line 238: also test with `description == null`

kulla (changes requested):

server/src/uuid/model/taxonomy_term/messages.rs
hugotiburtino: Alright, good. 
In legacy there is the [following](https://github.com/serlo/serlo.org/blob/main/packages/public/server/src/module/Taxonomy/src/Manager/TaxonomyManager.php#L137) though, but using a migration is even better.

How would we version the migration? Should we implement it in DB layer?

server/src/uuid/model/taxonomy_term/messages.rs
hugotiburtino: Since root will be done by migration, it is indeed unnecessary. Removing it...

server/src/uuid/model/taxonomy_term/messages.rs
hugotiburtino: I'm not sure if I undestood... which model.rs?

server/tests/taxonomy_term.rs
hugotiburtino: yes, it would. 
The only trade-off is that the test gets quite slow, and even more if we test description as null.

hugotiburtino:
@kulla ready to be re-reviewed. Please take a look at the discussions

server/src/uuid/model/taxonomy_term/messages.rs
kulla: Move enum `TaxonomyType` into `server/src/uuid/model/taxonomy_term`

server/src/uuid/model/taxonomy_term/messages.rs
kulla: > How would we version the migration? Should we implement it in DB layer?

Currently by https://github.com/serlo/serlo.org/tree/main/packages/migrations For shuting down legacy serlo.org we need to move this source code out. However this will be on of the last steps

server/tests/taxonomy_term.rs
```diff
@@ -219,3 +217,68 @@ mod move_mutation {
...
     }
 }
+
+#[cfg(test)]
+mod create_mutation {
+    use test_utils::*;
+
+    #[actix_rt::test]
+    async fn creates_new_taxonomy_term() {
+        for taxonomy_type in TAXONOMY_TYPES_WITHOUT_ROOT.iter() {
```
kulla, line 227:
```suggestion
        for taxonomy_type in ["topic", "topicFolder"].iter() {
```

That's enough.

test-utils/src/lib.rs
```diff
@@ -180,6 +184,25 @@ pub async fn set_entity_revision_field<'a>(
...
+    TaxonomyType::Topic,
+    TaxonomyType::TopicFolder,
+    TaxonomyType::Blog,
+    TaxonomyType::Curriculum,
+    TaxonomyType::CurriculumTopic,
+    TaxonomyType::CurriculumTopicFolder,
+    TaxonomyType::Forum,
+    TaxonomyType::ForumCategory,
+    TaxonomyType::Locale,
+];
```
kulla, line 204:
```suggestion
pub const TAXONOMY_TYPES_WITHOUT_ROOT: [TaxonomyType; 2] = [
    TaxonomyType::Topic,
    TaxonomyType::TopicFolder,
];
```

The others I do not want to support in the future...

kulla (approved):
@hugotiburtino Some small suggestions. Feel free to merge and to https://github.com/serlo/serlo.org-database-layer/pull/207#discussion_r852666450 in a follow up PR

server/tests/taxonomy_term.rs
hugotiburtino: since we have shortened the `TAXONOMY_TYPES_WITHOUT_ROOT`` (now called ALLOWED_TAXONOMY_TYPES_CREATE) there is no need to change this line

test-utils/src/lib.rs
hugotiburtino: Good. Changed, also the var name.
BTW: this is in test-utils/lib. But I suppose you would like to put it in production code isn't?

PR #208, opened 2022-04-14
Refactor TaxonomyTerm::move_parent()

kulla:


hugotiburtino (commented):
Just to understand: do you think that a positive check (`if let Some(previous_parent_id) = child.previous_parent_id.map(|x| x as i32)`) is better than some negative checks?

kulla:
This implementation omits an `.unwrap()` and is thus a little bit safer...

hugotiburtino (approved):

hugotiburtino:
> This implementation omits an .unwrap() and is thus a little bit safer...

I was also concerned about that, but since there was a `is_none()` check before, I thought it would be ok to use the unwrap() afterward.
Anyway, thanks for that.

Issue #210, opened 2022-04-14
TaxonomyCreateEntityLinkMutation


kulla:
### Message (TypeScript definition)

```ts
interface Message {
  type: "TaxonomyCreateEntityLinkMutation",
  payload: Payload
}

interface Payload {
  entityIds: number[]
  taxonomyTermId: number
}
```
### Output (TypeScript definition)

```ts
interface Output {
  success: boolean
}
```

## Description what happens in the database layer

Creates a new link in the table `term_taxonomy_entity` with the last of the possible positions.

## Necessary checks / possible errors

- [x] `entityIds` belong each to an entity
- [x] `taxonomyTermId` belongs to a taxonomy term
- [x] both elements (entity and taxonomy term) have the same instance
- [x] there is no link already between them
- [x] Create necessary events

PR #221, opened 2022-04-29
Add TaxonomyCreateEntityLinkMutation

hugotiburtino:
See #210


hugotiburtino:
- [ ] Should not run on CoursePage, GroupedExercise and Solution(?)

server/src/uuid/model/entity/mod.rs
```diff
@@ -490,9 +484,22 @@ impl Entity {
...
-}
 
-impl Entity {
+    pub async fn check_entity_exists<'a, E>(id: i32, executor: E) -> Result<(), operation::Error>
+    where
+        E: Executor<'a>,
+    {
+        if let Err(UuidError::DatabaseError {
+            inner: sqlx::Error::RowNotFound,
+        }) = Self::fetch_via_transaction(id, executor).await
```
kulla, line 494: This would result in `Ok(())` for all other `sqlx::Error` -> Try 

```rs
Self::fetch_via_transaction(...).await.map_err(....)
```

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -553,4 +554,100 @@ impl TaxonomyTerm {
...
+        executor: E,
+    ) -> Result<(), operation::Error>
+    where
+        E: Executor<'a>,
+    {
+        let mut transaction = executor.begin().await?;
+
+        let instance_id = Self::get_instance_id(payload.taxonomy_term_id, &mut transaction).await?;
+
+        for child_id in payload.entity_ids.clone() {
```
kulla, line 569:
```suggestion
        for child_id in payload.entity_ids {
```

Use `*child_id` instead of `child_id` then

server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -553,4 +554,100 @@ impl TaxonomyTerm {
...
+            .fetch_optional(&mut transaction)
+            .await?;
+
+            if is_child_already_linked_to_taxonomy.is_some() {
+                return Err(operation::Error::BadRequest {
+                    reason: format!(
+                        "Entity {} is already linked to taxonomy term {}",
+                        child_id, payload.taxonomy_term_id
+                    ),
+                });
```
kulla, line 590:
```suggestion
                continue;
```

server/tests/taxonomy_term.rs
```diff
@@ -282,3 +282,153 @@ mod create_mutation {
...
+
+        assert_bad_request(
+            response,
+            "Entity 28952 and taxonomy term 7 are not in the same instance",
+        )
+        .await;
+    }
+
+    #[actix_rt::test]
+    async fn fails_if_link_already_exists() {
```
kulla, line 411: Check that link is not stored twice.

kulla (changes requested):

server/src/uuid/model/entity/mod.rs
```diff
@@ -940,71 +947,6 @@ mod tests {
     async fn add_revision() {
         let pool = create_database_pool().await.unwrap();
         let mut transaction = pool.begin().await.unwrap();
-
-        Entity::add_revision(
```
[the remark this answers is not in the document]
hugotiburtino: We have meanwhile a way better [integration test](https://github.com/serlo/serlo.org-database-layer/blob/main/server/tests/entity.rs#L24) for that.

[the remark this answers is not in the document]
kulla: then you need to delete the whole test function...

server/tests/taxonomy_term.rs
kulla: No, it shouldn't throw an error...

hugotiburtino: check

kulla (approved):

PR #230, opened 2022-05-05
Fix autoreview entity create

kulla:
Fixes https://github.com/serlo/serlo.org-database-layer/issues/229

### DIFF
--- a/server/src/uuid/model/entity/mod.rs
+++ b/server/src/uuid/model/entity/mod.rs
@@ -442,7 +442,7 @@ impl Entity {
         .await?;
 
         if !payload.input.needs_review {
-            let _ = Entity::checkout_revision(
+            Entity::checkout_revision(
                 EntityCheckoutRevisionPayload {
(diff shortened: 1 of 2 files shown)


server/src/uuid/model/entity/mod.rs
```diff
@@ -441,16 +441,21 @@ impl Entity {
         .save(&mut transaction)
         .await?;
 
+        dbg!(payload.input.needs_review);
```
hugotiburtino, line 444:
```suggestion
```

hugotiburtino (changes requested):

kulla:
@hugotiburtino Resolved in https://github.com/serlo/serlo.org-database-layer/pull/230/commits/d6fad476624d7ac507ea23e31d80382d06c0e3a5

hugotiburtino (approved):
Fair enough.
Shouldn't we write test for that? I could do that in a followup issue

PR #240, opened 2022-05-09
DeletedEntitiesQuery

MoeHome:


server/src/uuid/model/entity/mod.rs
```diff
@@ -990,6 +995,64 @@ impl Entity {
...
+        let test = payload.after.as_ref();
+        match test {
+            Some(date) => {
+                let date_chrono = chrono::DateTime::parse_from_rfc3339(&(date))
+                    .unwrap()
+                    .with_timezone(&Utc);
+                date_database = Some(datetime::DateTime::from(date_chrono));
+            }
+            None => date_database = None,
+        }
```
hugotiburtino:
```suggestion
        let date_database: Option<datetime::DateTime> = match payload.after.as_ref() {
            Some(date) => {
                let date_chrono = chrono::DateTime::parse_from_rfc3339(&(date))
                    .unwrap()
                    .with_timezone(&Utc);
                Some(datetime::DateTime::from(date_chrono));
            }
            None => None,
        }
```

server/src/uuid/model/entity/messages.rs
```diff
@@ -302,3 +306,47 @@ impl MessageResponder for UnrevisedEntitiesQuery {
...
+
+pub mod deleted_entities_query {
+    use super::*;
+
+    #[derive(Deserialize, Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Payload {
+        pub first: i32,
+        pub after: Option<String>,
+        pub instance: Option<String>,
```
hugotiburtino, line 318: We have a type `Instance`, see example at `event/messages.rs`

server/src/uuid/model/entity/mod.rs
```diff
@@ -990,6 +995,64 @@ impl Entity {
...
```
hugotiburtino:
```suggestion
        let date_database: Option<DateTime> = match payload.after.as_ref() {
            Some(date) => {
                let date_chrono = chrono::DateTime::parse_from_rfc3339(&(date))
                    .unwrap()
                    .with_timezone(&Utc);
                date_database = Some(DateTime::from(date_chrono));
            }
            None => date_database = None,
        }
```

server/src/uuid/model/entity/mod.rs
```diff
@@ -19,14 +20,18 @@ use crate::event::{
     CreateEntityEventPayload, CreateEntityRevisionEventPayload, CreateTaxonomyLinkEventPayload,
     EntityLinkEventPayload, EventError, RevisionEventPayload,
 };
-use crate::{fetch_all_fields, format_alias};
+
+use crate::{datetime, fetch_all_fields, format_alias};
```
hugotiburtino, line 24:
```suggestion
use crate::{fetch_all_fields, format_alias};
```

server/src/uuid/model/entity/mod.rs
```diff
@@ -990,6 +995,64 @@ impl Entity {
...
+        let date_ref = payload.after.as_ref();
+        match date_ref {
+            Some(date) => {
+                let date_chrono = chrono::DateTime::parse_from_rfc3339(&(date))
+                    .unwrap()
+                    .with_timezone(&Utc);
+                date_database = Some(datetime::DateTime::from(date_chrono));
+            }
+            None => date_database = None,
+        }
```
hugotiburtino, line 1016:
```suggestion
        let date_database: Option<DateTime> = match payload.after.as_ref() {
            Some(date) => {
                let date_chrono = chrono::DateTime::parse_from_rfc3339(&(date))
                    .unwrap()
                    .with_timezone(&Utc);
                date_database = Some(DateTime::from(date_chrono));
            }
            None => date_database = None,
        }
```

server/src/uuid/model/entity/mod.rs
```diff
@@ -990,6 +995,64 @@ impl Entity {
...
+    ) -> Result<Vec<DeletedEntity>, sqlx::Error>
+    where
+        E: Executor<'a>,
+    {
+        let date_database: Option<DateTime>;
+        let test = payload.after.as_ref();
+        match test {
+            Some(date) => {
+                let date_chrono = chrono::DateTime::parse_from_rfc3339(&(date))
+                    .unwrap()
```
MoeHome, line 1011: @hugotiburtino  Shall I catch the error (wrong format) here?

server/src/uuid/model/entity/mod.rs
```diff
@@ -990,6 +995,64 @@ impl Entity {
...
+        executor: E,
+    ) -> Result<Vec<DeletedEntity>, sqlx::Error>
+    where
+        E: Executor<'a>,
+    {
+        let date_database: Option<DateTime>;
+        let date_ref = payload.after.as_ref();
+        match date_ref {
+            Some(date) => {
+                let date_chrono = chrono::DateTime::parse_from_rfc3339(&(date))
```
hugotiburtino, line 1010: on the other hand, it is not that good to parse it here, unwrapping an error could stop the application. Maybe the API could send a date as simple `number` , what do you think @kulla ?

hugotiburtino (changes requested):

server/src/uuid/model/entity/mod.rs
hugotiburtino: I was thinking about that, see https://github.com/serlo/serlo.org-database-layer/pull/240#discussion_r867970518

server/src/uuid/model/entity/mod.rs
MoeHome: Eventuell operation::Error::BadRequest zurückgeben?

hugotiburtino: Yes that is right

server/tests/entity.rs
```diff
@@ -371,3 +371,100 @@ mod create_mutation {
...
+            .await;
+
+        assert_ok_with(response, |result| {
+            assert_has_length(&result["deletedEntities"], first as usize);
+            assert_json_include!(
+                actual: &result["deletedEntities"][0],
+                expected: json!({
+                    "id": 14809,
+                    "dateOfDeletion": "2014-03-10T13:26:52+01:00",
+                })
```
hugotiburtino, line 394: Just a minor thing: `assert_json_include!` is to be used when we don't actually have to know what is going to come from the result (v.g. when creating an entity, we become automatically the field `date` filled in, which we cannot know beforehand what value it is going to get). For this use case though an `assert_json` would be enough like in https://github.com/serlo/serlo.org-database-layer/blob/main/server/tests/events.rs#L14-L24

It is not a bad thing to leave it as it is though.

server/src/uuid/model/entity/mod.rs
```diff
@@ -990,6 +994,68 @@ impl Entity {
...
+            date_database,
+            payload.instance,
+            payload.instance,
+            payload.first,
+        )
+        .fetch_all(executor)
+        .await?;
+
+        for entity in result {
+            let deletion_date: DateTime = entity.date.unwrap().into();
```
hugotiburtino, line 1048: I've tested and it works good. We are though uneasy with the use of `unwrap` here, since it could put the app in an unrecoverable state. See v.g. [this discussion](https://stackoverflow.com/questions/39477684/should-i-avoid-unwrap-in-production-application).

hugotiburtino, line 1048: I've tested and it works good. We are though uneasy with the use of `unwrap` here, since it could put the app in an unrecoverable state. See v.g. [this discussion](https://stackoverflow.com/questions/39477684/should-i-avoid-unwrap-in-production-application).

hugotiburtino (approved):

Issue #224, opened 2022-05-02
[New Endpoint] EntitySetLicenseMutation


kulla:
### Message (TypeScript definition)

```ts
interface Message {
  type: 'EntitySetLicenseMutation'
  payload: Payload
}

interface Payload {
  entityId: number!
  licenseId: number!
  userId: number!
}
```

### Output (TypeScript definition)

```ts
interface Output {
  success: boolean
}
```

## Description what happens in the database layer

Sets the license of entity with id `entityId` to `licenseId`

## Necessary checks / possible errors

- [ ] Does entity with license `entityId` exist?
- [x] Create SetLicense event

hugotiburtino:
Some TODOS: 
- [ ] Remove/drop from the database init.sql the `license` table and related constraints
Remember to do that also in staging and prod databases

MoeHome:
Should there be an info if the old license_id matches the new one?

hugotiburtino:
IMHO it doesn't have to have an extra info about that but it could  return successfully early.

MoeHome:
To specify: If an invalid user id isn't handled there will be an InternalServerError. We could change that to BadRequest.

kulla:
> To specify: If an invalid user id isn't handled there will be an InternalServerError. We could change that to BadRequest.

Would be cool if it will be a bad request

PR #249, opened 2022-05-12
add EntitySetLicenseMutation

MoeHome:


server/src/uuid/model/entity/mod.rs
```diff
@@ -1057,6 +1057,73 @@ impl Entity {
...
+            r#"
+                select * from user where id = ?
+            "#,
+            payload.user_id,
+        )
+        .fetch_optional(&mut transaction)
+        .await?
+        .ok_or(operation::Error::BadRequest {
+            reason: format!("An user with id {} does not exist.", payload.user_id),
+        })?;
```
hugotiburtino, line 1080: Indeed that is good for [this purpose](https://github.com/serlo/serlo.org-database-layer/issues/224#issuecomment-1124981732) but we should think later in a solution that changes the EventPayload. [here](https://github.com/serlo/serlo.org-database-layer/blob/main/server/src/event/model/event.rs#L162-L176)  we get a sql contraint error, that leads to the internal server error. You could try to catch and convert the error in this PR or we could open a refactor issue for that...

server/src/uuid/model/entity/mod.rs
```diff
@@ -1057,6 +1057,73 @@ impl Entity {
...
+
+        let instance_id: i32 = sqlx::query!(
+            r#"
+                select instance_id from entity where id = ?
+            "#,
+            payload.entity_id,
+        )
+        .fetch_one(&mut transaction)
+        .await?
+        .instance_id;
```
hugotiburtino, line 1116:
```suggestion
```
We have already this information from the query from the lines 1082-1092

server/src/uuid/model/entity/mod.rs
```diff
@@ -1057,6 +1057,73 @@ impl Entity {
...
+
+        CreateSetLicenseEventPayload::new(payload.entity_id, payload.user_id, instance_id)
```
hugotiburtino, line 1118:
```suggestion
        CreateSetLicenseEventPayload::new(payload.entity_id, payload.user_id, entity.instance_id)
```

server/tests/entity.rs
```diff
@@ -468,3 +468,124 @@ mod deleted_entities_query {
...
+        assert_eq!(new_license_id, license_id);
+
+        let last_event_log_entry = sqlx::query!(
+            r#"
+                select * from event_log order by date desc limit 1
+            "#,
+        )
+        .fetch_one(&mut transaction)
+        .await
+        .unwrap();
```
hugotiburtino, line 514: smart 😉
But it would be good the make a real integration test to see if the event notification was indeed created and can be fetched. [Example](https://github.com/serlo/serlo.org-database-layer/blob/main/server/tests/taxonomy_term.rs#L107-L125)

server/tests/entity.rs
```diff
@@ -468,3 +468,124 @@ mod deleted_entities_query {
...
+
+        let user_id: i32 = 1;
+        let entity_id: i32 = 0;
+        let license_id: i32 = 2;
+
+        let response = Message::new(
+            "SetLicenseMutation",
+            json!({"userId": user_id, "entityId": entity_id, "licenseId": license_id}),
+        )
+        .execute_on(&mut transaction)
```
hugotiburtino, line 533: You could try the suggestion of kulla (in Slack):
```suggestion
        .execute()
```
consequently, deleting line 523

server/tests/entity.rs
```diff
@@ -468,3 +468,124 @@ mod deleted_entities_query {
...
+
+        let user_id: i32 = 0;
+        let entity_id: i32 = 1495;
+        let license_id: i32 = 2;
+
+        let response = Message::new(
+            "SetLicenseMutation",
+            json!({"userId": user_id, "entityId": entity_id, "licenseId": license_id}),
+        )
+        .execute_on(&mut transaction)
```
hugotiburtino, line 555:
```suggestion
        .execute()
```

server/tests/entity.rs
```diff
@@ -468,3 +468,124 @@ mod deleted_entities_query {
...
+        .await;
+
+        let last_event_log_entry = sqlx::query!(
+            r#"
+                select uuid_id from event_log order by date desc limit 1
+            "#,
+        )
+        .fetch_one(&mut transaction)
+        .await
+        .unwrap();
```
hugotiburtino, line 587: Same as above about events. I'm just not sure how to negatively test...

hugotiburtino (changes requested):

server/src/uuid/model/entity/mod.rs
hugotiburtino: Let's do it afterwards. I've opened an issue for that #264

server/tests/entity.rs
hugotiburtino: So far in events we have just tested that the event in another one. If you can't find a better solution, you could keep this pattern...

hugotiburtino (approved):

Issue #178, opened 2022-03-14
Refactor: Submodularize some models


hugotiburtino:
Some modules have repeated code regarding payload, v.g. 
https://github.com/serlo/serlo.org-database-layer/blob/main/server/src/thread/messages.rs#L71
```
#[derive(Deserialize, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct ThreadCreateCommentMutation {
    pub thread_id: i32,
    pub content: String,
    pub user_id: i32,
    pub subscribe: bool,
    pub send_email: bool,
}
```
is the same as https://github.com/serlo/serlo.org-database-layer/blob/main/server/src/thread/model.rs#L145
```
#[derive(Debug, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct ThreadCommentThreadPayload {
    pub thread_id: i32,
    pub content: String,
    pub user_id: i32,
    pub subscribe: bool,
    pub send_email: bool,
}
```

It leads to less boilerplate and better overview if we  submodularize such payloads according to the already existent pattern in v.g. https://github.com/serlo/serlo.org-database-layer/blob/main/server/src/metadata/messages.rs#L31 and https://github.com/serlo/serlo.org-database-layer/blob/main/server/src/metadata/model.rs#L33

Besides that, some unit tests are unnecessary since too similar to the integration (or feature) test (in server/tests).

Here a list of the modules that would need such refactoring:
- [x] event
- [x] navigation
- [x] notification (@anbestCL )
- [x] thread
- [x] uuid/page 
- [x] uuid/entity

kulla:
@hugotiburtino You have assigned this issue to @anbestCL , right?

hugotiburtino:
Not completely. Anna would do some of them, I think she would like to start with `notification`

PR #252, opened 2022-05-13
Refactor event messages

kulla:
See #178


hugotiburtino:
You would need to refactor the tests, since `assert_bad_request` there does not require anymore the reason text.

hugotiburtino (approved):

PR #254, opened 2022-05-13
Refactor entity

### LINKED ISSUE #178: Refactor: Submodularize some models

kulla:
Fixes #178

### DIFF
--- a/server/src/uuid/model/entity/mod.rs
+++ b/server/src/uuid/model/entity/mod.rs
@@ -3,7 +3,7 @@ use async_trait::async_trait;
 use chrono::Utc;
 use convert_case::{Case, Casing};
 use futures::try_join;
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use sqlx::MySqlPool;
 use sqlx::Row;
 use std::collections::HashMap;
@@ -866,20 +866,12 @@ impl Entity {
     }
 }
 
-#[derive(Debug, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
-pub struct UnrevisedEntitiesQueryResult {
-    pub unrevised_entity_ids: Vec<i32>,
-}
-
 impl Entity {
-    pub async fn unrevised_entities<'a, E>(
-        executor: E,
-    ) -> Result<UnrevisedEntitiesQueryResult, sqlx::Error>
+    pub async fn unrevised_entities<'a, E>(executor: E) -> Result<Vec<i32>, sqlx::Error>
     where
         E: Executor<'a>,
     {
-        let unrevised_entity_ids = sqlx::query!(
+        Ok(sqlx::query!(
             r#"
                 SELECT
                     MIN(e.id) as entity_id,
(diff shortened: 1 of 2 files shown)


server/src/uuid/model/entity/mod.rs
```diff
@@ -913,7 +818,9 @@ impl Entity {
         }) = revision.concrete_uuid
         {
             if revision.trashed {
-                return Err(EntityRejectRevisionError::RevisionAlreadyRejected);
+                return Err(operation::Error::BadRequest {
```
hugotiburtino, line 821: See comment at line 791

server/src/uuid/model/entity/mod.rs
```diff
@@ -802,7 +753,9 @@ impl Entity {
             }) = repository.concrete_uuid
             {
                 if abstract_entity.current_revision_id == Some(revision_id) {
-                    return Err(EntityCheckoutRevisionError::RevisionAlreadyCheckedOut);
+                    return Err(operation::Error::BadRequest {
+                        reason: "revision is already checked out".to_string(),
```
hugotiburtino, line 757: We have been lately implementing in a way that such cases returns succesfully early in such cases. What do you think: keep the error or return ok?

server/src/uuid/model/entity/mod.rs
```diff
@@ -925,7 +832,9 @@ impl Entity {
             }) = repository.concrete_uuid
             {
                 if abstract_entity.current_revision_id == Some(revision_id) {
-                    return Err(EntityRejectRevisionError::RevisionCurrentlyCheckedOut);
+                    return Err(operation::Error::BadRequest {
+                        reason: "revision is checked out currently".to_string(),
```
hugotiburtino, line 836: see other comments.

hugotiburtino (changes requested):

hugotiburtino (approved):

Issue #262, opened 2022-05-17
Refactor: Reduce number of awaits in tests


kulla:
Currently we have two `await`s in our test code. Example: https://github.com/serlo/serlo.org-database-layer/blob/20cf25a2a6aa574ca2975b9fd2e5deba532d67f0/server/tests/user.rs#L38-L42 In the first we wait for the response and in the second we wait for reading the whole response body. In changing `MessageResult` (by adding also the body to this struct) we can reduce the number of awaits to one.

PR #265, opened 2022-05-17
Refactor: Reduce number of awaits in tests

kulla:
Fixes https://github.com/serlo/serlo.org-database-layer/issues/262


hugotiburtino (changes requested):
That is a quite strange [error in ci-cd](https://github.com/serlo/serlo.org-database-layer/runs/6479077673?check_suite_focus=true). It works locally. I'll merge it in another PR, that has also a strange ci-cd behavior and let's see

PR #266, opened 2022-05-17
Fix TaxonomyTermCreate: bind type_id to instance

hugotiburtino:


server/src/uuid/model/taxonomy_term/mod.rs
```diff
@@ -489,16 +489,25 @@ impl TaxonomyTerm {
...
-            r#"SELECT id FROM type WHERE name = ?"#,
-            payload.taxonomy_type
+            r#"
+                SELECT type.id FROM type
+                JOIN taxonomy
+                    ON taxonomy.type_id = type.id
+                JOIN instance
+                    ON taxonomy.instance_id = instance.id
+                WHERE type.name = ?
+                    AND instance_id = ?
```
kulla, line 502:
```suggestion
                SELECT type.id FROM type
                JOIN taxonomy
                    ON taxonomy.type_id = type.id
                WHERE type.name = ?
                    AND taxonomy.instance_id = ?
```

kulla (approved):
Only one small change...

PR #268, opened 2022-05-19
add PagesQuery and tests

MoeHome:


server/tests/page.rs
```diff
@@ -60,3 +60,35 @@ mod create_mutation {
         .await
     }
 }
+
+#[cfg(test)]
+mod pages_query {
+    use test_utils::*;
+
+    #[actix_rt::test]
+    async fn fetches_all_pages() {
```
hugotiburtino, line 69: I'm just wondering if we would like to keep the pattern of other tests, in which the whole response is tested.
See for example [tests/subject.rs](https://github.com/serlo/serlo.org-database-layer/blame/main/server/tests/subject.rs) (don't even look at navigation.rs 😆 )

server/tests/page.rs
```diff
@@ -60,3 +60,35 @@ mod create_mutation {
...
+        Message::new("PagesQuery", json!({}))
+            .execute()
+            .await
+            .should_be_ok_with(|result| assert_eq!(&result["pages"][0], 16256))
+            .await
+    }
+
+    #[actix_rt::test]
+    async fn fetches_all_pages_of_instance() {
```
hugotiburtino, line 78: See comment to previous test

hugotiburtino (approved):

PR #274, opened 2022-05-30
Filter some users

MoeHome:


hugotiburtino (changes requested):
Could you add a test for it?

Since in the test db there is no user with description, you could use a sqlx query to set description for some users

Issue #271, opened 2022-05-28
SetEntity… creates empty revision for parent


elbotho:
# What is wrong?

When editing an `ExerciseGroup` and only changing e.g. a solution of a `GroupedExercise` an empty revision of the `ExerciseGroup` is also created.
![image](https://user-images.githubusercontent.com/1258870/170816917-9fd8be25-355b-49a4-83b3-3faef2ba892e.png)

# How should it behave?

We should check if the parent changed before creating a new revision

kulla:
Okay, I found the reason: We added a new field `cohesive` which is read as `false` when it is not present in the DB. So when a `ExerciseGroup` is saved which current is so old that it's current revision does not have a field `cohesive` a new revision with `cohesive === false` is stored. I work on a fix.

PR #277, opened 2022-06-03
fix: Do not add new revision when only "cohesive" is added

kulla:
Closes https://github.com/serlo/serlo.org-database-layer/issues/271

### DIFF
--- a/server/src/uuid/model/entity/mod.rs
+++ b/server/src/uuid/model/entity/mod.rs
@@ -468,13 +468,21 @@ impl Entity {
         let mut fields = payload.input.fields.clone();
 
         if let Some(revision_id) = last_not_trashed_revision {
-            let last_revision_fields: HashMap<String, String> =
+            let mut last_revision_fields: HashMap<String, String> =
                 fetch_all_fields!(revision_id, &mut transaction)
                     .await?
                     .into_iter()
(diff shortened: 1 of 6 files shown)


server/src/uuid/model/entity/mod.rs
```diff
@@ -468,14 +468,22 @@ impl Entity {
...
-            let last_revision_fields: HashMap<String, String> =
+            let mut last_revision_fields: HashMap<String, String> =
                 fetch_all_fields!(revision_id, &mut transaction)
                     .await?
                     .into_iter()
                     .filter(|field| field.field != "changes")
                     .map(|field| (field.field.to_case(Case::Camel), field.value))
                     .collect();
 
+            // FIXME: This is bad design -> let's have a DB migration?!
```
hugotiburtino, line 479: > let's have a DB migration?!

At any rate! This cohesive implies a lot of extra code, opening the doors to bugs.
But what do you have in mind?

hugotiburtino (approved):

server/src/uuid/model/entity/mod.rs
kulla: Migration: Each exercise group revision without a `cohesive` property gets one with `cohesive === false`.

PR #282, opened 2022-06-07
feat: filter pages which do not have a revision

MoeHome:


hugotiburtino (approved):

server/src/uuid/model/page/mod.rs
```diff
@@ -490,9 +490,11 @@ impl Page {
         Ok(sqlx::query!(
             r#"
                 SELECT page_repository.id
-                FROM page_repository, instance
+                FROM instance, page_repository
+                JOIN page_revision ON page_repository.id = page_revision.page_repository_id
                 WHERE page_repository.instance_id = instance.id
                 AND (? is null or instance.subdomain = ?)
+                AND page_repository.current_revision_id = page_revision.id
```
kulla, line 497:
```suggestion
                AND page_repository.current_revision_id is not null
```

kulla (approved):
no blocker though but a possible improvement (haven't tested it)

server/src/uuid/model/page/mod.rs
MoeHome: That doesn't work, because there are pages in page_repository with a current_revision_id that is not theirs.
(the pages with ids 20310 and 20311 both have current_revision_id 16285, but in page_revision the revision with id 16285 has page_repository_id 16256.

Issue #285, opened 2022-06-07
For course pages a new revision is added when they have a legacy icon field


kulla:
# What is wrong?

Course pages sometimes have still the `icon` field from legacy which we do not support any more. Anytime a course is changed with revisions which still have the `icon` field a new revision is added.

# Steps to reproduce

1. `yarn mysql:import-anonymous-data`
2. Run

```
yarn fetch EntityAddRevisionMutation '{                                                                                   
  "revisionType": "CoursePageRevision",
  "userId": 1,
  "input": {
    "changes": "test",
    "entityId": 198935,
    "needsReview": true,
    "subscribeThis": true,
    "subscribeThisByEmail": true,
    "fields": {
[...]

PR #286, opened 2022-06-07
fix: Saving course page revisions without an icon

kulla:
Closes https://github.com/serlo/serlo.org-database-layer/issues/285

### DIFF
--- a/server/src/uuid/model/entity/mod.rs
+++ b/server/src/uuid/model/entity/mod.rs
@@ -484,6 +484,13 @@ impl Entity {
                 last_revision_fields.insert("cohesive".to_string(), "false".to_string());
             }
 
+            // FIXME: Do we need to fix this in the frontend?!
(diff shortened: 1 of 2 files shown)


server/src/uuid/model/entity/mod.rs
```diff
@@ -484,6 +484,13 @@ impl Entity {
                 last_revision_fields.insert("cohesive".to_string(), "false".to_string());
             }
 
+            // FIXME: Do we need to fix this in the frontend?!
```
hugotiburtino, line 487: I see this PR as a hotfix.
l actually think it is not a task of the DB Layer to like correct not supported fields.
But it is indeed easier to do it here than in the API (where it would be a more appropriate place).
Couldn't we plan also a migration for that?

hugotiburtino (approved):

PR #287, opened 2022-06-07
fix: Create AddRevisionEvent before CheckoutRevisionEvent

### LINKED ISSUE #270: Trigger `CreateEntityNotificationEvent` before `CheckoutRevisionNotificationEvent`
# What is wrong?

It seems currently in the event log it looks like entities that are created and accepted at the same time (by reviewers) are accepted before they are created.

kulla:
Closes #270


hugotiburtino:
Could you add a test for that?

hugotiburtino (changes requested):
Could you please add a test for that?

hugotiburtino:
> Could you please add a test for that?

done

hugotiburtino (approved):

PR #289, opened 2022-06-07
feat: TaxonomyTermSortMutation shall support subset of children ids

### LINKED ISSUE #288: taxonomyTerm.sort should work without all children ids
Originally from @elbotho 

# What happens?
Currently, taxonomyTerm.sort must receive all children, in order to sort them. But that makes life of frontend hard, since it has to fetch all children, even the trashed ones or entities without revision.

# Desired Behavior
taxonomyTerm.sort may receive just a slice of the current children ids and sort them, ignoring the rest.

kulla:
Closes #288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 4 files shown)


hugotiburtino (approved):
Thanks.
The only downside of this approach is that it doesn't throw an error if childrenIds is empty ([]), but we can live with that..,

PR #305, opened 2022-06-20
feat: add UserRemoveRoleMutation

MoeHome:


MoeHome:
That just came to my mind: Does this also have to check for inherited or passed to roles?

hugotiburtino (approved):
Very good, thanks, after your test in current system, we can think in merging it

MoeHome:
> > That just came to my mind: Does this also have to check for inherited or passed to roles?
> 
> First: could you see how does it behave in legacy?
> 
> Do you mean: if a user get a role A removed, but they still have role B, that inherits from A, so that we remove also role B (or throw an error, someting like 'user have inherited role A from role B, please remove B first')? If I got you right, that would be good from the perspective from the user, but since it is supposed to be a temporary endpoint (this is going to be managed in kratos later), I wouldn't spend much effort in it. Make it if you think it is easy, otherwise let's go on.

legacy only removes the specified role.

Yes, that was my thinking, but in my opinion, just removing is the best solution.

hugotiburtino (approved):

PR #306, opened 2022-06-22
 New Endpoint UserAddRoleMutation

MoeHome:


server/tests/user.rs
```diff
@@ -12,6 +12,84 @@ mod user_activity_by_type_query {
...
+            role_id
+        )
+        .fetch_all(&mut transaction)
+        .await
+        .unwrap();
+        assert_eq!(response[0].user_id as i32, user_id);
+    }
+
+    #[actix_rt::test]
+    async fn does_not_add_a_new_row() {
```
hugotiburtino, line 49:
```suggestion
    async fn does_not_add_a_new_row_if_user_already_has_role() {
```

hugotiburtino (changes requested):

hugotiburtino:
Very good, thanks, just a minor thing and we can merge it

PR #309, opened 2022-06-27
New Endpoint UserCreateMutation

MoeHome:


hugotiburtino:
Thanks, it seems good. I'll test it manually a little bit further before merging it.

PR #310, opened 2022-06-28
fix: Ignore user threads in AllThreadsQuery

### LINKED ISSUE #307: Ignore user comments on /discussions
See https://serlo.slack.com/archives/C0BMSC431/p1655900248776729 -> Since we only use user discussions for invitations we should filter them out on /discussions

kulla:
Fixes #307

### DIFF
--- a/server/src/thread/model.rs
+++ b/server/src/thread/model.rs
@@ -42,12 +42,14 @@ impl Threads {
                 SELECT comment.id
                 FROM comment
                 JOIN uuid ON uuid.id = comment.id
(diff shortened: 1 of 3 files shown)


hugotiburtino (approved):

hugotiburtino:
It seems good.
I just would like to be sure that @elbotho is aware of this change...

elbotho:
> 
It seems good.
I just would like to be sure that @elbotho is aware of this change...
If I don't hear any protest, I'll merge it 😄

Thanks, we discussed this.

elbotho (approved):

PR #337, opened 2022-10-26
feat: Order AllThreadsQuery according to last updated

hugotiburtino:
See https://serlo.slack.com/archives/C0BMSC431/p1666617395168269

### DIFF
--- a/server/src/thread/model.rs
+++ b/server/src/thread/model.rs
@@ -39,7 +39,7 @@ impl Threads {
 
         let result = sqlx::query!(
             r#"
-                SELECT comment.id
+                SELECT comment.id, comment.date
                 FROM comment
                 JOIN uuid ON uuid.id = comment.id
                 JOIN uuid parent_uuid ON parent_uuid.id = comment.uuid_id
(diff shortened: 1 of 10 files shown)


server/src/thread/model.rs
```diff
@@ -39,7 +39,7 @@ impl Threads {
 
         let result = sqlx::query!(
             r#"
-                SELECT comment.id
+                SELECT comment.id, comment.date
```
kulla, line 42: In our current understanding `comment.date` shall be the date when the comment was created and it shall not be updated for new comments

kulla: Use a SQL statement like `select comment.id, comment.date, MAX(answer.id) as most_recent_answer_id from comment join uuid on comment.uuid_id = uuid.id join comment answer on answer.parent_id = comment.id group by comment.id limit 10;` -> Order by `most_recent_answer_id` then

kulla (changes requested):

server/src/thread/model.rs
hugotiburtino: Thanks for your hint in slack 😃

PR #345, opened 2022-11-19
fix: Fix sorting of threads in discussions

hugotiburtino:
If ok, let me version it and merge.

### DIFF
--- a/server/src/uuid/model/entity/mod.rs
+++ b/server/src/uuid/model/entity/mod.rs
@@ -1,6 +1,5 @@
 use crate::uuid::Subject;
 use async_trait::async_trait;
-use chrono_tz::Europe::Berlin;
 use convert_case::{Case, Casing};
 use futures::try_join;
 use serde::Serialize;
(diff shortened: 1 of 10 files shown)


server/src/uuid/model/entity/mod.rs
```diff
@@ -918,19 +917,10 @@ impl Entity {
...
-                            .to_string(),
-                    }
-                })
-            })
-            .transpose()?
-            .map(|date| date.with_timezone(&Berlin).to_string());
+        let after_db_time = match payload.after.as_ref() {
+            Some(date) => DateTime::parse_from_rfc3339(date)?,
+            None => DateTime::now(),
+        };
```
kulla, line 923: Can be a helper function

kulla (approved):

server/src/uuid/model/entity/mod.rs
hugotiburtino: true, let's go on for now and we could do this refactoring if the need is bigger

PR #349, opened 2022-12-27
feat: Add `DeletedTaxonomiesQuery`

hugotiburtino:
The previously visible error in actions led to a more complicated fix, so I removed the commit from here, solved the error and explained it in #350


MoeHome (approved):

hugotiburtino:
As discussed in our meeting, we should prefer something like DeletedUuidsQuery, what demands a more abstract and complicated code.

PR #350, opened 2022-12-28
fix: Really trash or restore taxonomy term

hugotiburtino:
fix: Create event for trashing/restoring taxonomy term


hugotiburtino:
It is a little complicated to explain this fix, so let's do it slowly.

Until now, the table `term` was (mistakenly) used to fetch the taxonomy term id, but this table has its own id order. It means: if we asked to trash the uuid 1 it was found in the term table. After that it entered the check at https://github.com/serlo/database-layer/blob/3de492861a34bdefa1db92a3d4472419675d5078/server/src/uuid/model/uuid.rs#L297-L301, where it could throw an error in case the uuid were a user or entityRevision.

[ffc08e8](https://github.com/serlo/database-layer/pull/350/commits/ffc08e8fc4630a9098de24fcf3bbd1ac9ce61595) correctly uses the table `term_taxonomy`. It caused the error at ```left: `200`,  right: `400`', test-utils/src/lib.rs:82:9``` (see https://github.com/serlo/database-layer/actions/runs/3790223750/jobs/6444681991),  because now there were no match for the uuid 1. But it led to the check in https://github.com/serlo/database-layer/blob/main/server/src/uuid/model/uuid.rs#L323,  skipping the script and returning 200, although nothing happened.

Now, I excluded also the entity revisions at all from the query (see [be1b51d](https://github.com/serlo/database-layer/pull/350/commits/be1b51d7399c7eabbda72c1d6bbbb37c7faea1a2)) and made it throw an error in case the uuid is not found in the query (see [3de4928](https://github.com/serlo/database-layer/pull/350/commits/3de492861a34bdefa1db92a3d4472419675d5078)), either the uuid do not exist or it is a user, an entity revision or something else that cannot be trashed).

kulla (approved):

PR #357, opened 2023-01-05
feat: Add ThreadEditMutation

AndreasHuber:


server/src/thread/model.rs
```diff
@@ -350,6 +351,95 @@ impl Threads {
...
+            return Err(operation::Error::BadRequest {
+                reason: "content is empty".to_string(),
+            });
+        }
+
+        let mut transaction = executor.begin().await?;
+
+        let is_trashed = match sqlx::query!(
+            r#"
+                SELECT trashed FROM uuid WHERE id = ?
```
hugotiburtino, line 372: You can fetch this info in only one query, joining this query with the one in line 395.
use a join
``SELECT ...,, trashed FROM comment JOIN uuid ON uuid.id = comment.id ... `

server/src/thread/model.rs
```diff
@@ -350,6 +351,95 @@ impl Threads {
...
+            "#,
+            payload.comment_id
+        )
+        .fetch_one(&mut transaction)
+        .await
+        {
+            Ok(record) => record.trashed,
+            Err(_) => {
```
hugotiburtino, line 380: We could map exactly the error (it should be sqlx:Error:RowNotFound and throw it accordingly. The others error should be a server error.

server/src/thread/model.rs
```diff
@@ -350,6 +351,95 @@ impl Threads {
...
+    {
+        if payload.content.is_empty() {
+            return Err(operation::Error::BadRequest {
+                reason: "content is empty".to_string(),
+            });
+        }
+
+        let mut transaction = executor.begin().await?;
+
+        let is_trashed = match sqlx::query!(
```
hugotiburtino, line 370: Instead of using a `match` it is common to use `map_err` in this repo. 
I personally think it is easier to read with map_err

hugotiburtino (changes requested):

hugotiburtino:
thanks! Wonderful that you can already deliver a working code

AndreasHuber:
- Saving the edit date requires adding a column to the database table comment first, so this is missing or only the intent shown via commented code. 
- Touching the test cases for create comment / thread was not scope of the ticket but we had agreed I could do it here anyways.

AndreasHuber:
Question came up regarding naming convention. I thought EditCommentMutation would be a clear name but maybe the original suggestion ThreadUpdateMutation or something like ThreadEditCommentMutation is more consistent with existing naming. Please let me know and I can push an according rename. @kulla

server/src/thread/messages.rs
```diff
@@ -17,6 +17,7 @@ pub enum ThreadMessage {
     ThreadCreateThreadMutation(create_thread_mutation::Payload),
     ThreadCreateCommentMutation(create_comment_mutation::Payload),
     ThreadSetThreadArchivedMutation(set_thread_archived_mutation::Payload),
+    EditCommentMutation(edit_comment_mutation::Payload),
```
hugotiburtino, line 20:
```suggestion
    ThreadEditMutation(edit_comment_mutation::Payload),
```

server/src/thread/messages.rs
```diff
@@ -45,6 +46,9 @@ impl MessageResponder for ThreadMessage {
                     .handle("ThreadSetThreadArchivedMutation", connection)
                     .await
             }
+            ThreadMessage::EditCommentMutation(message) => {
```
hugotiburtino, line 49:
```suggestion
            ThreadMessage::ThreadEditMutation(message) => {
```

hugotiburtino (approved):
@kulla @MoeHome we are applying the moto: merge first, ask later. 
If you have change request, please open a new issue.

kulla:
> @kulla @MoeHome we are applying the moto: merge first, ask later.

:+1:

>  I thought EditCommentMutation would be a clear name but maybe the original suggestion ThreadUpdateMutation or something like ThreadEditCommentMutation

In the current convention the first subject is somethink like a namespace. The namespace `thread` is perfect, so I would go with `ThreadEditCommentMutation`

PR #367, opened 2023-02-22
Update dependencies

kulla:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.61 as build
+FROM rust:1.67 as build
```
kulla, line 1: @hugotiburtino Shall we add a .tool-versions file in this repo and make all changes with a pinned `rust` version?!

hugotiburtino: It sounds safer.
Since Rust is still envolving, I would just say that we should keep upgrading its version, in order to get the benefits of new language features (v.g. the last thing, that we got to know, was string interpolation). 
If we don't want to have the extra work of keeping upgrading it, I would say let's leave as it is.

kulla:
@hugotiburtino I started with the update but unfortunately there are pact errors

hugotiburtino:
> @hugotiburtino I started with the update but unfortunately there are pact errors

yes, quite ungrateful to debug. We may want to leave the Actions for rustup with tollchain 'stable' and just update rust version at `.tool-versions` when it breaks in the CI.

kulla:
I close it for now and reopen it when I work on it again :smile:

PR #374, opened 2023-03-31
WIP: Create metadata API according to AMB spec 

CodingDive:
Board: https://github.com/orgs/serlo/projects/35/views/1
Specification: https://dini-ag-kim.github.io/amb/draft


server/tests/metadata.rs
```diff
@@ -103,4 +103,128 @@ mod entities_metadata_query {
             .await
             .should_be_bad_request();
     }
+
+    #[actix_rt::test]
+    async fn checks_mandatory_keys_in_response() {
+        Message::new("EntitiesMetadataQuery", json!({ "first": 350 }))
```
kulla, line 109: No blocker but a follow up: I would make a fetch for an array of UUIDs each being of a different type we want support in the metadata API. Maybe a request like `returns_metadata_of_entities` but with different types is sufficient.

server/tests/metadata.rs
```diff
@@ -103,4 +103,128 @@ mod entities_metadata_query {
...
+                for mandatory_key in &mandatory_keys {
+                    assert!(
+                        entity.get(mandatory_key).is_some(),
+                        "Mandatory key '{}' is missing from the Quiz!",
+                        mandatory_key
+                    );
+                }
+            }
+        });
+    }
```
kulla, line 169:
```suggestion
```

When there is at least one exercise in the first 350 entities there is no additional value of this test than just having `default_value_for_property_name()` and `checks_mandatory_keys_in_response()`

kulla (changes requested):

PR #376, opened 2023-04-07
refactor: Proposal for easier structure of metadata

kulla:
In the past we had a file structure as 

```
src/submodule/
  |- mod.rs
  |- messages.rs
  |- model.rs
```

The idea behind this structure is that it enforces a separation between the definition of a message (i.e. the JSON we get) and an abstract model for our business logic in `model.rs`. In `messages.rs` should not include a SQL statement but instead those should be hidden behind a model in `model.rs` so that the functions of the model can be executed by many message calls. 

However let's take `EntitiesMetadataQuery`: The function in `src/metadata/model.rs` is only called by `src/metadata/messages.rs` and nowhere else. IMHO this will not change in the foreseeable future. Thus I would suggest in those cases to merge `messages.rs` and `model.rs` as in this PR -> Have a look at  https://github.com/serlo/database-layer/blob/refactor-metadata-structure/server/src/message.rs

What do you think about this refactoring?


hugotiburtino:
> What do you think about this refactoring?

I'm fan of not overengineering. So I would agree that we should at simple modules we should keep just one file.

My only worry is that we will have two approaches and it may be hard for some newbies to decide what approach to use.  So my opinion is: if you decide at the end of the day not to do this refactoring, I would be ok with it also.

AndreasHuber:
It looks like you did two things: 

1.  You created a metadata.rs file instead of a metadata/mod.rs. I think this change is the preferred way these days, see for example https://doc.rust-lang.org/rust-by-example/mod/split.html. With a mod.rs file is the old way. I am in favor of this change. 

2. You merged the code from the metadata/messages.rs and metadata/model.rs file into the module file, be it metadata/mod.rs or now metadata.rs. Looks fine to me, especially because I think the boilerplate in the messages.rs files can eventually be reduced for example by using the enum_dispatch crate. However, I have no strong opinion on this at the moment.

AndreasHuber (approved):

PR #380, opened 2023-04-15
refactor: Remove check of unique name at taxonomy term

hugotiburtino:


server/tests/taxonomy_term.rs
```diff
@@ -194,38 +164,6 @@ mod create_mutation {
         .await
         .should_be_bad_request();
     }
-
```
AndreasHuber, line 197: Do you want to remove the test cases or keep them and change the expectation?

AndreasHuber (approved):

server/tests/taxonomy_term.rs
hugotiburtino: Since we wouldn't test our code but the mysql database, I would just remove it. Someone in the future that sees this test would think, the logic is somewhere in code, but it isn't.

PR #393, opened 2023-04-26
test(metadata): Add test that query is fast enough

### LINKED ISSUE #392: Add test that EntityMetadataQuery is fast enough
Add test that EntityMetadataQuery querying 10.000 elements (somewhere in the middle of the current DB dump) succeeds fast (smaller than `X ms`). Find a good amount for `X`.

kulla:
Fixes #392

### DIFF
--- a/server/tests/metadata.rs
+++ b/server/tests/metadata.rs
@@ -321,10 +321,10 @@ mod entities_metadata_query {
             .unwrap()
             .as_millis();
 
-        // Querying 10.000 elements should be faster than 1 second, so that querying all entities
-        // will take less than 10 seconds (At April 2023 we had ~50.000 entities so even if we add
-        // taxonomies in the future it will be less than 100.000 objects).
-        assert!(end - start < 1000, "duration {:}", end - start);
+        // Querying 10.000 elements should be faster than 2 seconds, so that querying all entities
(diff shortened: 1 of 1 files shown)


server/tests/metadata.rs
```diff
@@ -302,6 +304,23 @@ mod entities_metadata_query {
         }));
     }
 
+    #[actix_rt::test]
+    async fn assert_that_query_is_faster_then_2000ms() {
```
AndreasHuber, line 308:
```suggestion
    async fn assert_query_is_faster_than_3000ms() {
```
Implementation of the test looks like 3s, not 2s.

AndreasHuber:
Tests for timing requirements can be a pain when they sometimes pass, sometimes fail. 
I think it's okay here but I would like to not have too many of this kind.

AndreasHuber (changes requested):

kulla:
> Tests for timing requirements can be a pain when they sometimes pass, sometimes fail. I think it's okay here but I would like to not have too many of this kind.

I am with you! (Currently this is the only case where I see an advantage having them). Thx for code review -> I have added your suggestions. Ready to be merged

AndreasHuber (approved):

PR #394, opened 2023-04-26
test: Add `pretty_assertions`

kulla:
Add
[`pretty_assertions`](https://docs.rs/pretty_assertions/latest/pretty_assertions) so that we get better diffs in the tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -550,6 +550,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctor"
+version = "0.1.26"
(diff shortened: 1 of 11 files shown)


AndreasHuber (approved):
Unfortunately still doesn't help with my most common question when looking at failed test cases: 
Which out of left/right is expectation/actual?

kulla:
> Unfortunately still doesn't help with my most common question when looking at failed test cases: Which out of left/right is expectation/actual?

Yep, I have added https://github.com/serlo/database-layer/issues/398 for this case

PR #403, opened 2023-04-28
fix(entity): prevent creation of two solutions for an exercise where the new one then is not shown

AndreasHuber:


server/src/uuid/model/entity/mod.rs
```diff
@@ -585,6 +586,28 @@ impl Entity {
     {
         let mut transaction = executor.begin().await?;
 
+        if payload.entity_type == EntityType::Solution
+            && sqlx::query!(
+                "SELECT uuid.id 
+                 FROM entity_link JOIN uuid ON uuid.id = entity_link.id
```
hugotiburtino, line 592:
```suggestion
                 FROM entity_link JOIN uuid ON uuid.id = entity_link.child_id
```

server/tests/entity.rs
```diff
@@ -186,107 +186,222 @@ mod add_revision_mutation {
...
+    #[case(EntityType::CoursePage, Option::<i32>::None, Some(18514))]
+    #[case(EntityType::Event, Some(7), Option::<i32>::None)]
+    #[case(EntityType::Exercise, Some(7), Option::<i32>::None)]
+    #[case(EntityType::ExerciseGroup, Some(7), Option::<i32>::None)]
+    #[case(EntityType::GroupedExercise, Option::<i32>::None, Some(2217))]
+    #[case(EntityType::Solution, Option::<i32>::None, Some(4827))]
+    #[case(EntityType::Video, Some(7), Option::<i32>::None)]
     #[actix_rt::test]
-    async fn creates_entity() {
-        for entity in EntityTestWrapper::all().iter() {
```
hugotiburtino, line 194: Finally we can get rid of this EntityTestWrapper. For that the parametrized tests are indeed useful. Please delete it also in the test utils.

server/src/uuid/model/entity/mod.rs
```diff
@@ -410,8 +410,9 @@ impl Entity {
                 SELECT c.id
                     FROM entity_link l
                     JOIN entity c on c.id = l.child_id
+                    JOIN uuid on uuid.id = c.id
                     JOIN type t ON t.id = c.type_id
-                    WHERE l.parent_id = ? AND t.name = ?
+                    WHERE l.parent_id = ? AND t.name = ? AND uuid.trashed = 0
```
hugotiburtino, line 415:
```suggestion
                    WHERE l.parent_id = ? AND t.name = ?
```
Since `fild_childen_by_id_and_type` are used by other cases, not just solution, we should keep it as it is.

I know it was in the issue, sorry for formulating there in a very misleading way. 
What I had in mind was to filter out trashed in `fild_childen_by_id_and_type` in `fild_child_by_id_and_type`, something like:
```rust
    async fn find_child_by_id_and_type (...) ...
    {
        Self::find_children_by_id_and_type(id, child_type, executor)
            .await
            .filter(|children| children.trashed == 0)
            .map(|children| children.first().cloned())
    }
```
For that you need of course to give this information in line 410 
```sql
SELECT c.id, trashed
```

hugotiburtino (changes requested):

server/tests/entity.rs
AndreasHuber: There are other tests still using EntityTestWrapper.

server/src/uuid/model/entity/mod.rs
AndreasHuber: Returning the trashed info might not result in the nicest code but I'll try and see.

server/src/uuid/model/entity/mod.rs
hugotiburtino: I think something went wrong when pulling, so that the original `entity_link.id` is back

AndreasHuber: Yes thanks, I must have accidentally overwritten the change with the text editor.

hugotiburtino (approved):

PR #412, opened 2023-05-08
feat(metadata): add field for metadata api mainentityofpage

AndreasHuber:


server/src/metadata.rs
```diff
@@ -251,6 +253,16 @@ pub mod entities_metadata_query {
                     is_family_friendly: true,
                     learning_resource_type: get_learning_resource_type(&result.resource_type),
                     license: LinkedNode { id: result.license_url},
+                    main_entity_of_page: json!({
+                        "id": "https://serlo.org/metadata-api",
+                        "provide": {
```
kulla, line 258:
```suggestion
                        "provider": {
```

server/src/metadata.rs
```diff
@@ -69,6 +69,7 @@ pub mod entities_metadata_query {
         is_part_of: Vec<LinkedNode>,
         learning_resource_type: Vec<LinkedNode>,
         license: LinkedNode,
+        main_entity_of_page: serde_json::Value,
```
kulla, line 72:
```suggestion
        main_entity_of_page: Value,
```

Isn't `Value` already in scope? By the way `serde_json::Value` might be better since it is more descriptive...

server/tests/metadata.rs
```diff
@@ -73,7 +101,22 @@ mod entities_metadata_query {
...
+        assert!(Utc::now() < actual_date_created + Duration::seconds(1));
+
+        let actual_date_modified = DateTime::parse_from_rfc3339(
+          &actual_response["entities"][0]["mainEntityOfPage"]["dateCreated"]
+            .as_str()
+            .unwrap()
+        ).unwrap();
+        assert!(Utc::now() < actual_date_modified + Duration::seconds(1));
+
+        let expected_response = json!({
```
kulla, line 119: Seems like something we can put in another function (I see duplicate code here) -> Can be a follow up issue though

kulla (changes requested):
Top :+1: Typo with `provider` is the only blocker though (the rest can be done in a follow up PR / issue)

server/src/metadata.rs
AndreasHuber: Value is not in scope. We could change that but yes serde_json::Value might be clearer anyway.

kulla (approved):

kulla:
Thx! Let's do https://github.com/serlo/database-layer/pull/412#discussion_r1187638036 as a follow up

PR #414, opened 2023-05-08
test(metadata): use function instead of copy paste

AndreasHuber:


server/tests/metadata.rs
```diff
@@ -1,28 +1,37 @@
...
-        assert!(Utc::now() < actual_date_created + Duration::seconds(1));
-
-        let actual_date_modified = DateTime::parse_from_rfc3339(
-          &actual_response["entities"][0]["mainEntityOfPage"]["dateCreated"]
-            .as_str()
-            .unwrap()
-        ).unwrap();
-        assert!(Utc::now() < actual_date_modified + Duration::seconds(1));
-
+        let (actual_date_created, actual_date_modified) = check_and_get_dates(&actual_response);
```
AndreasHuber, line 34: The indentation doesn't look right here and different than in similar cases below. 
It still looks the same after format though, any idea or tip @kulla ?

kulla (approved):

server/tests/metadata.rs
kulla: Yes, I have no idea why this actually happens -> in src/metadata.rs we have a similar problem. I think it might have to do with using macros...

kulla (approved):

kulla:
@AndreasHuber I have extended your solution with https://github.com/serlo/database-layer/pull/414/commits/f9b2820de387c70c497e144f64a4e57cf00fcdd0 What do you think about this?

server/tests/metadata.rs
AndreasHuber: Indeed, macro declarations and uses is mentioned under [rustfmt limiatations](https://github.com/rust-lang/rustfmt#limitations). Now we know.

PR #416, opened 2023-05-10
Add creator field

CodingDive:
Fixes https://github.com/serlo/database-layer/issues/384 and https://github.com/serlo/private-issues-sso-metadata-wallet/issues/35

### DIFF
--- a/server/src/metadata.rs
+++ b/server/src/metadata.rs
@@ -76,11 +76,6 @@ pub mod entities_metadata_query {
     }
 
     #[derive(Serialize)]
-    enum CreatorType {
-        Person,
-    }
-
-    #[derive(Serialize)]
     #[serde(rename_all = "camelCase")]
     struct Creator {
         #[serde(rename = "type")]
@@ -88,6 +83,11 @@ pub mod entities_metadata_query {
         id: String,
         name: String,
         affiliation: String,
+    }
+
+    #[derive(Serialize)]
+    enum CreatorType {
+        Person,
     }
 
     #[derive(Serialize)]
@@ -136,7 +136,8 @@ pub mod entities_metadata_query {
                     instance.subdomain AS instance,
                     JSON_ARRAYAGG(term_taxonomy.id) AS taxonomy_term_ids,
                     JSON_OBJECTAGG(term_taxonomy.id, term.name) AS term_names,
-                    JSON_OBJECTAGG(user.id, user.username) AS authors
+                    JSON_OBJECTAGG(user.id, user.username) AS authors,
+                    JSON_OBJECTAGG(all_revisions_of_entity.id, user.id) AS author_edits
                 FROM entity
                 JOIN uuid ON uuid.id = entity.id
                 JOIN instance ON entity.instance_id = instance.id
@@ -176,14 +177,24 @@ pub mod entities_metadata_query {
                     .map(|title| title.to_string());
                 let id = get_iri(result.id as i32);
 
-
                 let authors_map: HashMap<i32, String> = result.authors
                     .map_or_else(|| HashMap::new(), |authors| serde_json::from_value(authors).unwrap_or_default());
 
+                let edit_counts: HashMap<i32, usize> = result.author_edits
+                    .as_ref()
+                    .and_then(|edits| edits.as_object())
+                    .map(|edits| edits.values()
+                        .filter_map(|author_id| author_id.as_i64().map(|id| id as i32))
+                        .fold(HashMap::new(), |mut acc, author_id| {
+                            *acc.entry(author_id).or_insert(0) += 1;
+                            acc
+                        }))
+                    .unwrap_or_default();
 
                 let creators: Vec<Creator> = authors_map.iter()
-                    .sorted_by_key(|(id, _)| *id)
-                    .map(|(id,username)| Creator {
+                    .map(|(id, username)| (id, username, *edit_counts.get(id).unwrap_or(&0)))
+                    .sorted_by(|(id1, _, count1), (id2, _, count2)| count2.cmp(count1).then(id1.cmp(id2)))
+                    .map(|(id,username, _)| Creator {
                         creator_type: CreatorType::Person,
                         // Id is a url that links to our authors. It can look like
                         // the following
--- a/server/src/metadata.rs
+++ b/server/src/metadata.rs
@@ -49,26 +49,6 @@ pub mod entities_metadata_query {
     }
 
     #[derive(Serialize)]
-    enum CreatorType {
-        Person,
-        // Should we support this type too?
-        // Organization,
-    }
-
-    #[derive(Serialize)]
-    #[serde(rename_all = "camelCase")]
-    struct Creator {
-        #[serde(rename = "type")]
-        creator_type: CreatorType,
-        id: Option<String>,
-        name: String,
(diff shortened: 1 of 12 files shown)


server/src/metadata.rs
```diff
@@ -48,6 +48,26 @@ pub mod entities_metadata_query {
...
+    #[serde(rename_all = "camelCase")]
+    struct Creator {
+        #[serde(rename = "type")]
+        creator_type: CreatorType,
+        id: Option<String>,
+        name: String,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        affiliation: Option<serde_json::Value>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        honorific_prefix: Option<String>,
```
kulla, line 68:
```suggestion
```

Since we do and will not support it, I would this -> less code :smile:

server/src/metadata.rs
```diff
@@ -48,6 +48,26 @@ pub mod entities_metadata_query {
...
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    struct Creator {
+        #[serde(rename = "type")]
+        creator_type: CreatorType,
+        id: Option<String>,
+        name: String,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        affiliation: Option<serde_json::Value>,
```
kulla, line 66:
```suggestion
        affiliation: Option<serde_json::Value>,
```

`affiliation` can be Serlo:

```json
{
  "affiliation": {
    "id": "https://serlo.org/",
    "type": "Organization",
    "name": "Serlo Education e.V."
  }
}
```

(JSON should be the same as in `provider`)

server/src/metadata.rs
```diff
@@ -58,6 +78,8 @@ pub mod entities_metadata_query {
         schema_type: Vec<String>,
         date_created: String,
         date_modified: String,
+        // The authors of the resource
```
kulla, line 81:
```suggestion
```

See https://github.com/labs42io/clean-code-typescript#comments

server/src/metadata.rs
```diff
@@ -48,6 +48,26 @@ pub mod entities_metadata_query {
         entities: Vec<EntityMetadata>,
     }
 
+    #[derive(Serialize)]
+    enum CreatorType {
+        Person,
+        // Should we support this type too?
+        // Organization,
```
kulla, line 55:
```suggestion
        // Should we support this type too?
        // Organization,
```

No

server/src/metadata.rs
```diff
@@ -48,6 +48,26 @@ pub mod entities_metadata_query {
         entities: Vec<EntityMetadata>,
     }
 
+    #[derive(Serialize)]
+    enum CreatorType {
+        Person,
```
kulla, line 53: I have the feeling that an enum for just one value is not needed here (since currently this enum is only used at one place)... However it's not a big deal to leave the enum here, I do not have strong options.

server/src/metadata.rs
```diff
@@ -48,6 +48,26 @@ pub mod entities_metadata_query {
...
+        // Should we support this type too?
+        // Organization,
+    }
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    struct Creator {
+        #[serde(rename = "type")]
+        creator_type: CreatorType,
+        id: Option<String>,
```
kulla, line 63:
```suggestion
        id: String,
```

We will never have `None` here.

server/src/metadata.rs
```diff
@@ -155,6 +180,24 @@ pub mod entities_metadata_query {
...
+                        .unwrap_or_default();
+
+                let creators: Vec<Creator> = authors_map.iter()
+                    .map(|(id, username)| Creator {
+                        creator_type: CreatorType::Person,
+                        // Id is a url that links to our authors. It looks like
+                        // the following
+                        // https://serlo.org/user/:userId/:username e.g
+                        // https://serlo.org/user/240298/felix_eccardt
+                        id: Some(format!("https://serlo.org/user/{}/{}", id, username)),
```
kulla, line 195:
```suggestion
                        id: get_iri(id),
```

Actually `https:://serlo.org/<user-id>` works -> Having the username in the ID is also not a good idea since it might change (we sometimes rename users).

server/src/metadata.rs
```diff
@@ -155,6 +180,24 @@ pub mod entities_metadata_query {
...
+                let authors_map: HashMap<String, String> = serde_json::from_value(result.authors.unwrap_or_default())
+                        .unwrap_or_default();
+
+                let creators: Vec<Creator> = authors_map.iter()
+                    .map(|(id, username)| Creator {
+                        creator_type: CreatorType::Person,
+                        // Id is a url that links to our authors. It looks like
+                        // the following
+                        // https://serlo.org/user/:userId/:username e.g
+                        // https://serlo.org/user/240298/felix_eccardt
```
kulla, line 194:
```suggestion
```

server/src/metadata.rs
```diff
@@ -48,6 +48,26 @@ pub mod entities_metadata_query {
...
+    #[derive(Serialize)]
+    enum CreatorType {
+        Person,
+        // Should we support this type too?
+        // Organization,
+    }
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    struct Creator {
```
kulla, line 60: Put this type after struct EntityMetadata, see https://github.com/serlo/documentation/wiki/Best-Practices#order-source-files-from-abstract-to-concrete

server/src/metadata.rs
```diff
@@ -155,6 +180,24 @@ pub mod entities_metadata_query {
                     .and_then(|title| title.as_str())
                     .map(|title| title.to_string());
                 let id = get_iri(result.id as i32);
+
+
+                let authors_map: HashMap<String, String> = serde_json::from_value(result.authors.unwrap_or_default())
+                        .unwrap_or_default();
```
kulla, line 186: Just an idea:

```suggestion
                let authors_map: HashMap<String, String> = result.authors.and_then(|authors| serde_json::from_value(authors)).unwrap_or_default();
```

server/tests/metadata.rs
```diff
@@ -24,6 +24,53 @@ mod entities_metadata_query {
                     "LearningResource",
                     "Article"
                   ],
+                  "creator": [
```
kulla, line 27: To fix the tests: Make the order deterministic (e.g. Order by username alphabetically or order by user.id)

test-utils/src/lib.rs
```diff
@@ -125,16 +127,18 @@ where
         .await?
         .id as i32;
 
+    let mut rng: StdRng = Seeder::from(SEED).make_rng();
```
kulla, line 130: Should not be needed, see above

kulla (changes requested):

kulla (changes requested):

kulla (approved):

server/src/metadata.rs
kulla: > I changed the comment but left it in as for me id is badly chosen key by amb, it's not clear to me that a URL is behind this.

By the way: This does not come from AMB but from JSON-LD -> see the `@id` keyword there.

PR #430, opened 2023-05-23
feat: filter threads

MoeHome:


hugotiburtino:
We know already that a CTE that just uses the term_taxonomy table would not contemplate the comments for entities that have entities as parent. 
Indeed if we run the current script against the staging dump, ignoring all ?'s, we get as result 392.

Now, if we do
```sql
SELECT comment.id
                 FROM comment
                 JOIN uuid ON uuid.id = comment.id
                 JOIN comment answer ON comment.id = answer.parent_id  OR comment.id = answer.id
                 JOIN uuid parent_uuid ON parent_uuid.id = comment.uuid_id
                 -- put here a filter -- 
                 WHERE  comment.uuid_id IS NOT NULL
                     AND uuid.trashed = 0
                     AND comment.archived = 0
                     AND  parent_uuid.discriminator != "user"
                  GROUP BY comment.id;
```
we get 3674 results. That is the maximum allThreads would give back.

We could now investigate how they are distributed by replacing '-- put here a filter --' with different JOIN's.
- join entity_link on child_id = comment.uuid: 2264
- join term_taxonomy_entity on entity_id = comment.uuid: 999
- join term_taxonomy on term_taxonomy.id = comment.uuid: 392 (same as with cte, as expected)
- join page_repository on page_repository.id = comment.uuid_id: 19
Total: 3674

How to contemplate them in the CTE?

MoeHome:
The AllThreadsQuery doesn't look for threads on pages anymore. It would add more complexity to the code and there are only 58 of those threads. The last one is from 2020 and I think commenting pages was disabled at some point.

hugotiburtino (approved):

PR #431, opened 2023-05-24
Add about

MoeHome:


server/src/metadata.rs
```diff
@@ -291,8 +305,20 @@ pub mod entities_metadata_query {
...
+                            .filter_map(|element| element.as_i64())
+                            .collect::<HashSet<i64>>()
+                            .into_iter()
+                            .map(|id| id as i32)
+                            .collect()
+                    })
+                    .unwrap_or(Vec::new());
 
                 EntityMetadata {
+                    about: subject_ids,
```
kulla, line 321: See https://dini-ag-kim.github.io/amb/draft/#about for the definition. The values must be from https://skohub.io/dini-ag-kim/schulfaecher/heads/main/w3id.org/kim/schulfaecher/index.de.html

kulla (changes requested):

server/src/metadata.rs
MoeHome: Should the subjects of other instances also receive the German name?

MoeHome: And how shall I procede with subjects which do not have an equivalent subject in the schema, e.g. "Nachhaltigkeit" could be "Nature Conservation" but it's not the same or "Community" which doesn't have any similar subject?

kulla: > Should the subjects of other instances also receive the German name?

I don't understand you. The values are from a vocabulary and language agnostic. See https://dini-ag-kim.github.io/amb/draft/schemas/about.json for the formal definition and https://github.com/dini-ag-kim/amb/blob/main/draft/examples/valid/highered-course.json#L22 for how they should be added (ignore `prefLabel` here).

kulla: > And how shall I procede with subjects which do not have an equivalent subject in the schema, e.g. "Nachhaltigkeit" could be "Nature Conservation" but it's not the same or "Community" which doesn't have any similar subject?

I can give you an answer but since you are the product owner of this feature: What do you suggest?

kulla:
@MoeHome Maybe add a branch first to add support for mysql 8.x in this DB so that all tests pass. this can than be a basis for this PR and your other PR.

server/src/metadata.rs
MoeHome: My suggestion: If it is close enough to a course, we can point to it, otherwise nullify as @hugotiburtino suggested.

kulla: > My suggestion: If it is close enough to a course, we can point to it, otherwise nullify as @hugotiburtino suggested.

:+1: (here `nullify` means omit it in the result, right? Then yes!)

server/src/metadata.rs
```diff
@@ -388,4 +425,153 @@ pub mod entities_metadata_query {
...
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    struct SubjectMetadata {
+        r#type: String,
+        id: String,
+        in_scheme: String,
+    }
+
+    struct RawSubjectMetadata {
+        name: String,
```
kulla, line 462:
```suggestion
```

server/src/metadata.rs
```diff
@@ -388,4 +425,153 @@ pub mod entities_metadata_query {
...
+
+    #[derive(Serialize)]
+    #[serde(rename_all = "camelCase")]
+    struct SubjectMetadata {
+        r#type: String,
+        id: String,
+        in_scheme: String,
+    }
+
+    struct RawSubjectMetadata {
```
kulla, line 461: I would define a function which converts a `RawSubjectMetadata` into a `SubjectMetadata` -> See https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

server/src/metadata.rs
```diff
@@ -291,8 +306,30 @@ pub mod entities_metadata_query {
...
+                            .map(|id| id as i32)
+                            .collect()
+                    })
+                    .unwrap_or(Vec::new());
+                let subject_metadata: Option<Vec<SubjectMetadata>> = subject_ids.iter().map(|id| {
+                    let raw_subject_metadata = lookup_subject_metadata(*id);
+                    return if let Some(ref _meta) = raw_subject_metadata {
+                        Some(SubjectMetadata {
+                            r#type: "Concept".to_string(),
+                            id: raw_subject_metadata.as_ref().unwrap().in_scheme.to_id_string() + &raw_subject_metadata.as_ref().unwrap().id,
```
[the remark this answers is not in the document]
MoeHome: Then [this](https://github.com/serlo/database-layer/blob/ea84f0db8092c6b571e1b3a1240d8d80f8680024/server/src/metadata.rs#L339) shouldn't be there either?

hugotiburtino (approved):

server/src/metadata.rs
```diff
@@ -145,8 +147,19 @@ pub mod entities_metadata_query {
 
         Ok(sqlx::query!(
             r#"
+                WITH RECURSIVE ancestors AS (
+                    SELECT id, parent_id, id AS origin_id, id as subject_id
```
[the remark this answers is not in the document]
MoeHome: parent_id, id and subject_ id represent three layers of hierarchy. The query chooses the entries where parent_id is NULL and then the subjects are the grandchildren of NULL. The origin_id is the id of the taxonomy that the subject_id is assigned to.

PR #449, opened 2023-06-20
440 sync local db with production db

AndreasHuber:
change to local database:
ALTER TABLE notification MODIFY email TINYINT(1) NOT NULL DEFAULT '0', MODIFY email_sent TINYINT(1) NOT NULL DEFAULT '0';


AndreasHuber:
It seems to me like database should be changed correctly now, see lines 327649 and 327650. So I have to figure out why the Github Actions fail with the same error as if the database was unchanged.

docker-entrypoint-initdb.d/001-init.sql
```diff
@@ -327640,18 +327640,18 @@ UNLOCK TABLES;
...
-/*!40101 SET character_set_client = utf8 */;
+/*!50503 SET character_set_client = utf8mb4 */;
 CREATE TABLE `notification` (
-  `id` int(11) NOT NULL AUTO_INCREMENT,
-  `user_id` bigint(20) NOT NULL,
+  `id` int NOT NULL AUTO_INCREMENT,
+  `user_id` bigint NOT NULL,
   `seen` tinyint(1) NOT NULL DEFAULT '0',
   `date` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `email` tinyint(1) NOT NULL,
```
AndreasHuber, line 327649: still looking intact after changing MySQL version

hugotiburtino: The problem was at first place that the sqlx.json wasn't updated. But we also added a default value, as it is in production.

Issue #451, opened 2023-06-20
Investigate speed of metadata API


kulla:
See https://serlo.slack.com/archives/C04TTH20JP5/p1687177170415059

eliflores:
Closed thanks to https://github.com/serlo/database-layer/pull/452 🎉

PR #452, opened 2023-06-20
fix(metadata): Use fields directly instead of JSON_OBJECTAGG()

### LINKED ISSUE #450: Error when using metadata API in production
# What is wrong?

The following query resolves in an error after we updated the metadata API:

```
query {
  metadata {
    resources(first: 1, after: "MTYwOQ==") {
      pageInfo {endCursor}
    }
  }
}
```

The cursor `MTYwOQ==` belongs to UUID https://serlo.org/1609 -> So there is an error when accessing the metadata of the UUID https://serlo.org/1611. We get the following error in the DB layer:

```
EntitiesMetadataQuery: InternalServerError: ColumnDecode { index: "2", source: Utf8Error { valid_up_to: 5897, error_len: Some(1) } }
```

Description of [ColumnDecode](https://docs.rs/sqlx/latest/sqlx/error/enum.Error.html#variant.ColumnDecode): "Error occurred while decoding a value from a specific column."

kulla:
Fixes #450


kulla:
By accident we also solved #451 : We decreased the performance by 90% :tada:

hugotiburtino (approved):

PR #467, opened 2023-06-27
chore(Github Actions): let Dependabot create PRs vor version upgrades

AndreasHuber:


.github/dependabot.yml
```diff
@@ -0,0 +1,34 @@
...
+updates:
+  # repetitive because unfortunately Dependabot ticket for wildcard support still open:
+  # https://github.com/dependabot/dependabot-core/issues/2178
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+
+  - package-ecosystem: 'cargo'
+    directory: '/'
```
AndreasHuber, line 13: If Dependabot runs for example cargo upgrade in the given directory, then listing the subdirectories would not be necessary. But I am not sure what Dependabot does exactly, so we may find we need to add the subdirectories.

.github/workflows/checks.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - main
   pull_request:
+  merge_group:
```
AndreasHuber, line 7: after this change you can activate merge queue @kulla @hugotiburtino

PR #478, opened 2023-06-28
chore(dependabot): Add `@` to `types/node`

eliflores:
In order to ignore major versions of `@types/node`.

Follow up of https://github.com/serlo/database-layer/pull/474.


eliflores:
closing because we tried that [here](https://github.com/serlo/db-migrations), and dependabot did not close the PR to  bump @types/node from 18.16.13 to 20.3.2 💔

Issue #342, opened 2022-11-14
Refactor: Fix unwrap() in datetime.rs


kulla:
After #341 we introduced a `unwrap()` in https://github.com/serlo/database-layer/blob/main/server/src/datetime.rs#L50-L57 -> It would be great to have better error handling here.

PR #500, opened 2023-07-09
#342 custom error message in datetime.rs

iMac7:


AndreasHuber:
I guess that with better error handling @kulla meant a solution that handles the Option so the function cannot panic. Using expect can panic just like unwrap, only that it shows the added message when it panics.

iMac7:
Appreciate the suggestion, I'll work on it

PR #509, opened 2023-07-12
Upgrade sqlx to 0.7.1

AndreasHuber:
This PR includes: 
- Upgrade of sqlx
- Changes to make database-layer and tests compile again

Things to look out for when reviewing: 
- In a few places connection use has been changed to transaction use because using sqlx::Acquire::acquire in fetch leads to higher-ranked lifetime error that with sqlx::Acquire::begin is not there. This should not be a problem for performance but it is important that no transaction commit is forgotten resulting in a change not being saved to the database. The default without commit is rollback. Test cases should catch this but it is possible we have not everything covered by test cases.

Things that could make sense but are excluded here because the PR is already big:
- Entirely removing the self-made Executor trait
- Refactoring for execute functions that are not doing much anymore

Related info: 
- The [PR removing the Executor implementations for Transaction and PoolConnection ](https://github.com/launchbadge/sqlx/pull/2039), reason for most of the changes here
- How to write a method that can [take either a Transaction or a PgPool](https://github.com/launchbadge/sqlx/issues/970)
- Switch from [sqlx-data.json to .sqlx/](https://github.com/launchbadge/sqlx/commit/3f2ec6f91aa87db9f9f15adec1027b4126cde5f8) for builds without active database connection


server/src/message.rs
```diff
@@ -43,22 +45,25 @@ pub enum Message {
 #[async_trait]
 impl MessageResponder for Message {
     #[allow(clippy::async_yields_async)]
-    async fn handle(&self, connection: Connection<'_, '_>) -> HttpResponse {
+    async fn handle<'e, A: sqlx::Acquire<'e, Database = sqlx::MySql> + std::marker::Send>(
+        &self,
+        acquire_from: A,
+    ) -> HttpResponse {
         match self {
```
AndreasHuber, line 52: There's a crate to get rid of this match monstrosity.

server/src/alias/model.rs
```diff
@@ -111,14 +107,14 @@ where
...
             .await?
             .map(|result| result.uuid_id as i32)
             .ok_or(operation::Error::NotFoundError)?
         }
     };
 
-    let uuid = Uuid::fetch_via_transaction(id, &mut transaction).await?;
+    let uuid = Uuid::fetch_via_transaction(id, &mut *transaction).await?;
 
     transaction.commit().await?;
```
AndreasHuber, line 119: Is there actually something to commit to the database?

server/src/navigation/messages.rs
```diff
@@ -39,14 +41,12 @@ pub mod navigation_query {
...
+            acquire_from: A,
+        ) -> operation::Result<Self::Output> {
             let instance = self.instance.clone();
-            Ok(match connection {
-                Connection::Pool(pool) => Navigation::fetch(instance, pool).await?,
-                Connection::Transaction(transaction) => {
-                    Navigation::fetch_via_transaction(instance, transaction).await?
-                }
-            })
+            Ok(Navigation::fetch_via_transaction(instance, acquire_from).await?)
```
AndreasHuber, line 49: Not so much left of those execute implementations. Could possibly be scratched all together at some point.

server/src/subscription/model.rs
```diff
@@ -45,15 +45,16 @@ impl Subscriptions {
...
 
-    pub async fn fetch_by_object<'a, E>(object_id: i32, executor: E) -> Result<Self, sqlx::Error>
-    where
-        E: Executor<'a>,
-    {
+    pub async fn fetch_by_object<'a, A: sqlx::Acquire<'a, Database = sqlx::MySql>>(
+        object_id: i32,
+        acquire_from: A,
+    ) -> Result<Self, sqlx::Error> {
+        let mut transaction = acquire_from.begin().await?;
```
AndreasHuber, line 52: With acquire() instead of begin() this won't compile (higher-ranked lifetime error). I think a transaction is fine here but currently lack understanding why it doesn't work with a connection.

server/src/message.rs
AndreasHuber: out of scope, might do that in another PR later

hugotiburtino (approved):
Congrats and thanks a lot. Let's set a new version for the server in this PR in order to deploy it soon.

server/src/alias/messages.rs
```diff
@@ -47,15 +49,13 @@ pub mod alias_query {
...
+            acquire_from: A,
+        ) -> operation::Result<Self::Output> {
             let path = self.path.as_str();
             let instance = self.instance.clone();
-            Ok(match connection {
-                Connection::Pool(pool) => fetch(path, instance, pool).await?,
-                Connection::Transaction(transaction) => {
-                    fetch_via_transaction(path, instance, transaction).await?
-                }
-            })
```
kulla, line 58: Nice to not have those lines any more :-)

server/src/alias/model.rs
kulla: doesn't hurt either? Might be also faster since there is no rollback in the mysql...

server/src/database.rs
```diff
@@ -31,8 +31,8 @@ impl<'a, A> Acquire<'a> for A where A: sqlx::Acquire<'a, Database = sqlx::MySql>
 ///     E: Executor<'a>,
 /// {
 ///     let mut transaction = executor.begin().await?;
-///     let events = sqlx::query!(r#"SELECT id FROM event_log"#).fetch_all(&mut transaction).await?;
-///     let users = sqlx::query!(r#"SELECT id FROM user"#).fetch_all(&mut transaction).await?;
+///     let events = sqlx::query!(r#"SELECT id FROM event_log"#).fetch_all(&mut *transaction).await?;
+///     let users = sqlx::query!(r#"SELECT id FROM user"#).fetch_all(&mut *transaction).await?;
```
kulla, line 35: Do we still need this file?!

kulla (approved):

kulla:
> Congrats and thanks a lot. Let's set a new version for the server in this PR in order to deploy it soon.

@AndreasHuber Can you do this in another PR?

server/src/database.rs
AndreasHuber: Yes because I have not entirely removed its usage yet, only where it was necessary to make database-layer compile again.

PR #526, opened 2023-07-25
Refactor import script by putting logic in functions and warn if db in container is not running

CodingDive:
Refactored the logic a little bit after struggling yesterday to get the import script to run on the computer of Anna. It took us one hour to figure out that the docker container with the database was not running :facepalm:

### DIFF
--- a/scripts/mysql/mysql-import-anonymous-data.ts
+++ b/scripts/mysql/mysql-import-anonymous-data.ts
@@ -1,45 +1,88 @@
 import { spawnSync } from 'child_process'
 
-const latestDump = spawnSync(
-  'bash',
-  [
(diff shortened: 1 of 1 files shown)


eliflores:
Hi @CodingDive 👋 

This script is generated via the [/configure-repositories](https://github.com/serlo/configure-repositories) repository - may I ask you to move the change and open a PR there, instead, please? [This](https://github.com/serlo/configure-repositories/blob/main/files/scripts/mysql/mysql-import-anonymous-data.ts) is the file in that repository.
We could then re-generate the file here, and maybe to other repositories that could benefit from this change.

Thank you!

kulla (approved):

PR #531, opened 2023-07-27
Remove some boilerplate

AndreasHuber:
There is a lot of repetitive match stuff that would be nice to throw out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -789,6 +789,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "enum_dispatch"
+version = "0.3.12"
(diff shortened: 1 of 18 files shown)


AndreasHuber:
I was able to simplify some things but to get rid of the big chunk that I had in mind, I am not sure if the enum_dispatch and async_trait macros can be made to work together for this code, see https://gitlab.com/antonok/enum_dispatch/-/issues/42. 
Using the native async fn in traits would be nice but it is not yet available in stable Rust, only in nightly: https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#fn5

kulla (approved):

PR #610, opened 2023-11-06
feat: Exclude unsupported entites from unrevised entities

kulla:


server/src/uuid/model/entity/mod.rs
```diff
@@ -883,10 +883,16 @@ impl Entity {
...
+                JOIN type ON e.type_id = type.id
                 JOIN uuid u_e ON e.id = u_e.id
                 WHERE ( e.current_revision_id IS NULL OR r.id > e.current_revision_id )
                     AND u_r.trashed = 0
                     AND u_e.trashed = 0
+                    AND type.name NOT IN ("input-expression-equal-match-challenge",
+                        "input-number-exact-match-challenge", "input-string-normalized-match-challenge",
+                        "math-puzzle", "multiple-choice-right-answer", "multiple-choice-wrong-answer",
+                        "single-choice-right-answer", "single-choice-wrong-answer",
+                        "text-solution")
```
elbotho, line 895: Pretty sure things like `input-number-exact-match-challenge` did not end up in the review before.
Why did you add it here?

kulla: > Pretty sure things like input-number-exact-match-challenge did not end up in the review before. Why did you add it here?

Should increase performance a little bit in case there are unrevised revisions for those entities in the DB -> the API does not need to do the filtering any more...

server/src/uuid/model/entity/mod.rs
```diff
@@ -883,10 +883,16 @@ impl Entity {
                 FROM entity_revision r
                 JOIN uuid u_r ON r.id = u_r.id
                 JOIN entity e ON e.id = r.repository_id
+                JOIN type ON e.type_id = type.id
                 JOIN uuid u_e ON e.id = u_e.id
                 WHERE ( e.current_revision_id IS NULL OR r.id > e.current_revision_id )
                     AND u_r.trashed = 0
                     AND u_e.trashed = 0
+                    AND type.name NOT IN ("input-expression-equal-match-challenge",
```
hugotiburtino, line 891: It is more intuitive to put the supported revisions (instead of a negative statement). So
```sql
AND type.name IN ("article", "course"...)
```

hugotiburtino (changes requested):

server/src/uuid/model/entity/mod.rs
hugotiburtino: it is not a bad thing though, let's merge

hugotiburtino (approved):

PR #634, opened 2023-12-04
test: move integration tests into one common subfolder so that rustc …

AndreasHuber:
…builds them as one binary. Inspired by https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html. 

Checks/test is now taking 2:55 min (https://github.com/serlo/database-layer/actions/runs/7085501130/job/19281861240?pr=634) compared to around 3:50 if you look into other runs (like https://github.com/serlo/database-layer/actions/runs/7053591194/job/19200915517), so roughly 1 min less it seems.


kulla (approved):
Thank you!

PR #684, opened 2024-01-30
Remove user set description endpoint

AndreasHuber:
after the functionality is moved to the API gateway with https://github.com/serlo/api.serlo.org/pull/1234


hugotiburtino:
I would just leave the endpoints in database layer. 
It is an unnecessary task to remove it everytime we move the corresponding logic to the api

PR #721, opened 2024-03-22
refactor!: Delete AllThreadsQuery

kulla:
This was moved to API


kulla:
We need to wait until `staging` in API is deployed to `production`

PR #738, opened 2024-04-14
test: Remove pact tests

kulla:
* ~`test_pacts` needs to be removed from necessary CI commands in `main` in the GitHub settings of this repo~, see https://github.com/serlo/database-layer/pull/738#issuecomment-2053970811
* See https://github.com/serlo/api.serlo.org/issues/1363


kulla:
Info: I already removed the pact test as a necessary CI step so that I can merge other PRs => we need to enable it again when we think the pact tests are still necessary.

AndreasHuber (approved):

